PR #1, opened 2016-07-17
Completed C# Compiler

LogicAndTrick:
I started this a few days ago, and I just noticed that you were working on it as well so hopefully this can save you some time. I've compiled the test cases and the outputs all build using .NET Core. I haven't tried running them, though.

I've made some changes so that the result fits into the .NET naming standards and it takes advantage of some of the C#/.NET features that Java doesn't have.

- Added a .NET namespace config option (but defaults to `Kaitai` if not specified)
- Using .NET naming standards (public members use `PascalCase`, private members use `camelCase`)
    - Properties can't have the same name as classes, so the public getters are named `Get{Thing}()` instead of just `{Thing}()`
- I've done a (hopefully) complete implementation of the `KaitaiStream` class in my runtime fork at [LogicAndTrick/kaitai_struct_csharp_runtime](https://github.com/LogicAndTrick/kaitai_struct_csharp_runtime), let me know if you would like a PR for the runtime.
    - The runtime stream is based off .NET's `BinaryReader` class which has implementations for little-endian types, so only the BE methods need to be implemented.
    - As a bonus the `BinaryReader` already has `ReadSingle` and `ReadDouble` (and others), so floating point support comes free.
- All the classes are marked as `partial` so users can add methods and properties to the class without touching the generated file

Apologies if this conflicts with anything you're working on! I've never used Scala before, so hopefully I didn't do anything wrong.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
@@ -3,40 +3,50 @@ package io.kaitai.struct.languages
 import io.kaitai.struct.{LanguageOutputWriter, RuntimeConfig, Utils}
 import io.kaitai.struct.exprlang.Ast
 import io.kaitai.struct.exprlang.Ast.expr
-import io.kaitai.struct.exprlang.DataType.{BytesType, CalcIntType, StrType, _}
-import io.kaitai.struct.format.{NoRepeat, RepeatEos, RepeatExpr, RepeatSpec, _}
-import io.kaitai.struct.translators.{BaseTranslator, JavaTranslator, TypeProvider}
+import io.kaitai.struct.exprlang.DataType._
+import io.kaitai.struct.format._
+import io.kaitai.struct.translators.{BaseTranslator, CSharpTranslator, TypeProvider}
 
-class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
+class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter, namespace: String = "")
   extends LanguageCompiler(verbose, out)
     with EveryReadIsExpression
     with NoNeedForFullClassPath {
   import CSharpCompiler._
 
-  override def getTranslator(tp: TypeProvider): BaseTranslator = new JavaTranslator(tp)
+  override def getTranslator(tp: TypeProvider): BaseTranslator = new CSharpTranslator(tp)
 
   override def fileHeader(topClassName: String): Unit = {
     out.puts(s"// $headerComment")
-    out.puts
-    out.puts("using System;")
-    out.puts("using Kaitai;")
+
+    var ns = "Kaitai";
+    if (!namespace.isEmpty) ns = namespace;
 
     out.puts
+    out.puts("using System;")
+    out.puts("using System.Collections.Generic;")
+    if (ns != "Kaitai") out.puts("using Kaitai;")
+    out.puts
+
+    out.puts(s"namespace $ns")
+    out.puts(s"{")
+    out.inc
+  }
+
+  override def fileFooter(topClassName: String): Unit = {
+    out.dec
+    out.puts("}")
   }
 
   override def classHeader(name: String): Unit = {
-    val staticStr = if (out.indentLevel > 0) {
-      "static "
-    } else {
-      ""
-    }
 
-    out.puts(s"public ${staticStr}class ${type2class(name)} : $kstructName {")
+    out.puts(s"public partial class ${type2class(name)} : KaitaiStruct")
+    out.puts(s"{")
     out.inc
 
-    out.puts(s"public static ${type2class(name)} fromFile(String fileName) {")
+    out.puts(s"public static ${type2class(name)} FromFile(string fileName)")
+    out.puts(s"{")
     out.inc
-    out.puts(s"return new ${type2class(name)}(new $kstreamName(fileName));")
+    out.puts(s"return new ${type2class(name)}(new KaitaiStream(fileName));")
     out.dec
     out.puts("}")
   }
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -3,40 +3,50 @@ package io.kaitai.struct.languages
...
+    out.puts(s"public partial class ${type2class(name)} : KaitaiStruct")
+    out.puts(s"{")
     out.inc
 
-    out.puts(s"public static ${type2class(name)} fromFile(String fileName) {")
+    out.puts(s"public static ${type2class(name)} FromFile(string fileName)")
+    out.puts(s"{")
     out.inc
-    out.puts(s"return new ${type2class(name)}(new $kstreamName(fileName));")
+    out.puts(s"return new ${type2class(name)}(new KaitaiStream(fileName));")
```
GreyCat: What's the idea of changing it back to literal KaitaiStream?

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -48,72 +58,99 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
+    out.puts("_parent = parent;")
+    out.puts("_root = root;")
+    out.puts("ParseInternal();")
+    out.dec
+    out.puts("}")
+    out.puts
+
+    out.puts("private void ParseInternal()")
+    out.puts("{")
+    out.inc
```
GreyCat: Basically, all that mess with 3 constructors in Java was only because Java does not support default arguments for constructors. There's no need to reimplement in for C#, as far as I understand.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -48,72 +58,99 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
+  override def classConstructorFooter: Unit = {
+    out.dec
+    out.puts("}")
+    out.puts
+  }
 
   override def attributeDeclaration(attrName: String, attrType: BaseType, condSpec: ConditionalSpec): Unit = {
-    out.puts(s"private ${kaitaiType2NativeType(attrType)} ${privateMemberName(attrName)};")
+    out.puts(s"private ${kaitaiType2NativeType(attrType)} ${camelCase(attrName)};")
   }
```
GreyCat: Basically, the whole idea behind having `privateMemberName` is that you can fix naming style whenever you want, changing it in only one place. There's no point in changing it to literally "camelCase", "pascalCase", "upperCamelCase" or whatever's there in utilities — in future we might provide a switch for people to control it in runtime.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -48,72 +58,99 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
 
   override def attributeReader(attrName: String, attrType: BaseType): Unit = {
-    out.puts(s"public ${kaitaiType2NativeType(attrType)} ${publicMemberName(attrName)}() { return ${privateMemberName(attrName)}; }")
+    out.puts(s"public ${kaitaiType2NativeType(attrType)} Get${pascalCase(attrName)}() { return ${camelCase(attrName)}; }")
   }
```
GreyCat: I guess the "proper" C# way would be to have a named read-only property that will provide access to the internal private member.

The [code style we've chosen so far](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) mandates having private member names like `_FooBar` and public properties names like `FooBar`. However, `_xxx` is a syntax reserved for internal KS members that have special meaning and shouldn't clash with user-made stuff (i.e. `_root`, `_io`, `_parent`). In this case we're allowed to choose an alternative style - i.e. `mFooBar` for internal members and `FooBar` for public accessors.

I don't really see the point in using `GetFooBar` style. At the very least, we *can* have a clash here, if someone would introduce both `foo_bar` and `get_foo_bar` fields.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
LogicAndTrick: I think this may have been caused because I started from the Java compiler originally. I must have missed a few changes when I was matching it back to what you had. There's a couple of others with the renaming of the camelCase/member naming functions, I'll get them cleaned up.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
LogicAndTrick: I had a bunch of issues with variable naming, it felt a bit fragile trying to get everything to fit together because I had to change references in several places even though they're printing the same variable name. This combination is the only way I could make it work in the end. I'll play around with it more and see if I can get it working with the underscore prefixes and get it all back to the standard for internal fields.

I tried using read-only properties for getters to start with, but due to the way sub-objects are stored as inner classes, I got a compile error on the "position_in_seq" test case. `Header` is both a class and a property name, so it won't compile. The only way to solve the clash is to name it differently, which is why I put `Get` in front. But you're right about the clashes. Ideally the sub-objects wouldn't be inner classes, but I couldn't work out a way to change that behaviour. I could use non-standard naming instead (e.g. camelCase) but that's unappealing. I'm not sure what the answer is here, any ideas?

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -48,72 +58,99 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
 
   override def attrFixedContentsParse(attrName: String, contents: Array[Byte]): Unit = {
-    out.puts(s"${privateMemberName(attrName)} = _io.ensureFixedContents(${contents.length}, new byte[] { ${contents.mkString(", ")} });")
+    out.puts(s"${camelCase(attrName)} = _stream.EnsureFixedContents(${contents.length}, new byte[] { ${contents.mkString(", ")} });")
```
GreyCat: Actually, I'd like to keep naming of KaitaiStream instance associated with the KaitaiStruct with a name based off `_io`, to be consistent with all other runtimes. In case of C#, `_io` should translate into `m_Io`, I guess - but, again, that should be probably just generated with a call like `privateMemberName("_io")`.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -48,72 +58,99 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
   }
 
   override def attrProcess(proc: ProcessExpr, varSrc: String, varDest: String): Unit = {
     proc match {
       case ProcessXor(xorValue) =>
-        out.puts(s"this.$varDest = _io.processXorInt(this.$varSrc, ${expression(xorValue)});")
+        out.puts(s"this.$varDest = _stream.ProcessXorInt(this.$varSrc, ${expression(xorValue)});")
       case ProcessZlib =>
-        out.puts(s"this.$varDest = _io.processZlib(this.$varSrc);")
+        out.puts(s"$varDest = _stream.ProcessZlib($varSrc);")
```
GreyCat: Actually, all this `this.*` stuff should be replaced with `privateMemberName(*)` call.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -127,14 +164,14 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
 
   override def condRepeatEosHeader(id: String, io: String, dataType: BaseType, needRaw: Boolean): Unit = {
     if (needRaw)
-      out.puts(s"this._raw_${lowerCamelCase(id)} = new ArrayList<byte[]>();")
-    out.puts(s"this.${lowerCamelCase(id)} = new ${kaitaiType2NativeType(ArrayType(dataType))}();")
-    out.puts(s"while (!$io.isEof()) {")
+      out.puts(s"_raw_${id} = new List<byte[]>();")
```
GreyCat: That's actually a thing I wanted to have an in-depth discussion with the rest of the people interested in C# support. KS internal type system usually mandates one "all-around" sequential collection, as most script languages do, i.e. something that vaguely corresponds to notion of "array" in Ruby, Python, PHP and JS, and that's what `ArrayList` is generally accepted to be in Java.

Does `List` in C# fulfill the same role? I.e. a good all-around general sequential collection, with reasonably fast appends, not terribly huge memory overhead, etc?

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -160,187 +197,143 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
   override def handleAssignmentSimple(id: String, expr: String): Unit = {
-    out.puts(s"${privateMemberName(id)} = $expr;")
+    out.puts(s"${camelCase(id)} = $expr;")
   }
 
   override def parseExpr(dataType: BaseType, io: String): String = {
     dataType match {
       case t: IntType =>
-        s"$io.read${Utils.capitalize(t.apiCall)}()"
+        s"$io.Read${Utils.capitalize(t.apiCall)}()"
```
GreyCat: That is a good idea, actually :)

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -160,187 +197,143 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
       // Aw, crap, can't use interpolated strings here: https://issues.scala-lang.org/browse/SI-6476
       case StrByteLimitType(bs, encoding) =>
-        s"$io.readStrByteLimit(${expression(bs)}, " + '"' + encoding + "\")"
+        s"$io.ReadStringByteLimit(${expression(bs)}, " + '"' + encoding + "\")"
```
GreyCat: Can't really agree on these renames, though. Again, it's better to keep them all consistent in all runtimes.

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -160,187 +197,143 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
+    out.dec
   }
 
   override def instanceAttrName(instName: String): String = instName
 
-  override def instanceFooter: Unit = classConstructorFooter
+  override def instanceFooter: Unit = {
+    out.dec
+    out.puts(s"}")
+  }
```
GreyCat: Why would you actually want to inline all these?

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
```diff
@@ -160,187 +197,143 @@ class CSharpCompiler(verbose: Boolean, out: LanguageOutputWriter)
...
-    case _ => Utils.upperCamelCase(name)
-  }
+  override def privateMemberName(ksName: String): String = s"${Utils.lowerCamelCase(ksName)}"
 }
 
 object CSharpCompiler extends LanguageCompilerStatic with UpperCamelCaseClasses {
   override def indent: String = "    "
   override def outFileName(topClassName: String): String = s"${type2class(topClassName)}.cs"
+  override def outFilePath(config: RuntimeConfig, outDir: String, topClassName: String): String =
+    s"$outDir/src/${config.dotNetNamespace.replace('.', '/')}/${outFileName(topClassName)}"
```
GreyCat: I'm not sure, but from what I've seen so far, it seems that .NET does not have mandatory path components for packages, as Java does. And especially it does *not* have that `src` dir, as used by maven / some other build tools.

GreyCat:
Thank you so much for you hard work here, it's certainly really helpful :)

I've reviewed and commented all the changes you've sent. Could you fix / respond to stuff I've mentioned and update this pull request, so I could apply it right away?

I'm currently working on a testing system for C#, and actually it's almost finished. I'll try to deploy it to Travis soon enough, so we can get things rolling in C# department, with proper CI and progress measurements at http://kaitai.io/ci/

As a side note, a little advice for the future commits: could you please send them in more distinct, separate commits, so I could have applied some right away?

shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
LogicAndTrick: Yeah, `List<T>` is .NET's generic collection class, the equivalent of Java's `ArrayList<T>`. Their implementations are very similar (array-based, size doubling).

LogicAndTrick:
No problems, thanks for the quick review. I'll get those changes in as soon as I can.

GreyCat:
> I've done a (hopefully) complete implementation of the KaitaiStream class in my runtime fork at LogicAndTrick/kaitai_struct_csharp_runtime, let me know if you would like a PR for the runtime.

I've already reviewed and I agree with most of it, except for `GetXXX` stuff and renaming of `_io` to stream (`_io` should probably become `m_Io`, as per MS / .NET core codestyle). But I'm definitely in no position to judge any C# code, you people obviously know it much better than I do. Please do the PR, so other people interested in the C# support could review it too (i.e. @indrora).

LogicAndTrick:
I'm working on this now, the naming of things is going to get a bit messy because of the conflicts. Java allows for something like this:

```
class Test {
  private String example;
  public String example() { return example; }
}
```

Whereas this will result in a compiler error in C# due to duplicate declarations of the same name. I'm going with this alternative for now:

```
class Test {
  private String _example;
  public String example { get { return _example; } }
}
```

This doesn't suit the naming standard (public property should be PascalCase), but then it breaks in the case of an inner class...

```
class Test {
  public class Example { ... }
  private Example _example;
  public Example Example { get { return _example; } } // <- `Example` is already defined in class
}
```

The other option is to name inner classes something else, such as `_Example` or `Example_Inner`. But that will also deviate from the standard. Which do you think would be the better option? Break C# naming standards by using camelCase for getters, or change the name of inner classes?

Or should I use PascalCase and leave the test cases broken to be fixed later?

LogicAndTrick:
Closing this - Created PR #2 instead.

PR #2, opened 2016-07-18
Completed C# Compiler - 2nd Attempt

LogicAndTrick:
I've made changes from the review in #1 and rebased it onto the original `csharp` branch. I've split the changes into multiple commits, hopefully that makes it easier to review. Let me know if I've done anything wrong!

--- 
There is one remaining issue which is the naming conflicts I explained in the previous PR.

Right now I've gone with the "standards first" approach, which basically favours the C# naming standards over passing the test cases.

Five of the test cases generate classes that do not compile. One option to fix it is to use camelCase for the public getters instead of CamelCase, but this deviates from the standard naming convention that's used in .NET. I've made a patch and a branch with an extra commit to make this change and fix the test cases. The branch is [csharp-camelcase](https://github.com/LogicAndTrick/kaitai_struct_compiler/tree/csharp-camelcase) and the diff is [here](https://github.com/LogicAndTrick/kaitai_struct_compiler/blob/csharp-camelcase/Use%20camelCase%20for%20public%20getters.diff). Let me know and I'll add it to this PR branch.

### DIFF
--- a/jvm/src/main/scala/io/kaitai/struct/Main.scala
+++ b/jvm/src/main/scala/io/kaitai/struct/Main.scala
@@ -13,8 +13,9 @@ object Main {
 
     private val _verbose: Boolean = false,
     private val _debug: Boolean = false,
-    private val _javaPackage: String = ""
-  ) extends RuntimeConfig(_verbose, _debug, _javaPackage)
+    private val _javaPackage: String = "",
+    private val _dotNetNamespace: String = ""
+  ) extends RuntimeConfig(_verbose, _debug, _javaPackage, _dotNetNamespace)
 
   val ALL_LANGS = LanguageCompilerStatic.NAME_TO_CLASS.keySet
   val VALID_LANGS = ALL_LANGS + "all"
@@ -54,6 +55,10 @@ object Main {
       opt[String]("java-package") valueName("<package>") action { (x, c) =>
         c.copy(_javaPackage = x)
       } text("Java package (Java only, default: root package)")
+
+      opt[String]("dotnet-namespace") valueName("<namespace>") action { (x, c) =>
+        c.copy(_dotNetNamespace = x)
+      } text(".NET Namespace (.NET only, default: Kaitai)")
 
       opt[Unit]("verbose") action { (x, c) =>
         c.copy(_verbose = true)
--- a/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
@@ -255,7 +255,7 @@ object ClassCompiler {
   }
 
   private def getCompiler(lang: LanguageCompilerStatic, config: RuntimeConfig, out: LanguageOutputWriter) = lang match {
-    case CSharpCompiler => new CSharpCompiler(config.verbose, out)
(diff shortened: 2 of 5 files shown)


GreyCat:
Thanks for your huge work :) I'm reviewing it now.

Let me elaborate on the naming style issue.

First of all, this problem is *not* unique to C#, it's also exactly the same in C++ (although it arises when we're favoring STL-like style, with lowercase class and member names — it is "solved" there by appending `_t` to class names, i.e. member `header` vs class `member_t`). It's kind of the same in JS, although in JS it is somewhat alleviated by language using distinct styles for class vs member names.

Second, actually there is a similar problem that arises *not* only because of class vs member naming clash, but also in other cases. For example, note that different languages have different keywords (which may or may not be available to be used as identifiers). It's ok to use an attribute named `when` in Java, but it clashes with keyword `when` in Ruby.

For the time being, I suggest that we take the following stance on the problem:

1. KS does *not* deal with such language incompatibilities. Actually, it's quite practical: if you want something to compile for targets A, B and C, then you have to choose identifiers in a way that will be ok to use in both targets A, B and C. It's not like a showstopper - you can always rename them in a way that will satisfy all targets. May be some day, if we'll want to spend lots of time and effort to solve this, we'll review this policy and make some kind of universal identifier substitution/mangling solution, but, again, at this level it should be solved in some uniform way for every such problem. For example, if we encounter some identifier that will be invalid in a given language, we just mangle it by appending `_` suffix, so Ruby will get `when_` for field name and C# will get `Header_` for class name and `Header` for member name (or vice versa, doesn't really matter right now).

2. We try to follow language's native practices and standards, at least with default compilation options. This also means that if there's no universal solution for class vs member naming problem in a certain language, we don't try to make it up for language designers and insist on some kind of a solution (by default). We might (or might not, for some time being) provide alternative options, but ultimately it's up to user to give it names & compilation options that will be ok to his/her project.

3. For those who want different code generation options, we will provide alternative tweaks that. For example, allows one to have C# code with `lowerCamelCase` methods + `UpperCamelCase` classes, or even something more radical, like prefixed or suffixed classes (i.e. `KSUpperCamelCase` or `UpperCamelCaseInternal`, as you've suggested) or public member names (i.e. `GetUpperCamelCase`, as you proposed). But it's not a top priority task, and it is *definitely* optional.

4. For our own tests, I suggest that we just fix them all to avoid this problem. It's not such a big deal to fix five .ksy files to make it comfortable to work with C# (and also help C++ too).

GreyCat:
Phew! C# tests are finally up: http://kaitai.io/ci/

Have to fix broken Java now, and we'll be ready to merge!

LogicAndTrick:
Looks good!

Thanks for the details on the naming stuff, I think that's a good way to handle it. Changing the test cases will let the unit tests run, that's a good way to fix it for now.

Eventually, I think each compiler should know about potential conflicts and try to work around them if possible. In the case of C# it would just be the inner class - I think renaming the class (e.g. `class Header_ { ... }`) is the better option, as users will very rarely (almost never) refer to the inner class by name, rather grabbing data from the property instead. So I think the property should have the priority in that case.

It might be useful to allow template authors to specify their own language specific overrides in the template itself? Something like this:

```
seq:
  - id: when
    type: when
    name:
      - ruby: when_value
types:
  when:
    seq:
      - id: value
        type: u4
    name:
      - csharp: WhenInstance
```

PR #3, opened 2016-07-24
Add floating point data type

LogicAndTrick:
This adds a base data type for floating point values. There's a sample test case in a branch here: https://github.com/kaitai-io/kaitai_struct_tests/commit/613a01bd20cd6c00bb6bf9633558dbcd6244168c

- I'm using the `IntWidth` type, but should I create a new `FloatWidth` type instead? Or perhaps `IntWidth` should be renamed to something more generic like `TypeWidth`?
- The `FloatType` extends the `BaseType`, but it shares an interface with `IntType`, so it might be better to abstract the `apiCall` method into a parent type `ApiType` or `NativeType`? I'm not sure what the best way to approach it is in Scala.

Because they don't share an interface, the compiler can't look like this:
```
      case t: IntType | t: FloatType => // or just `case t: ApiType`
        s"$io.Read${Utils.capitalize(t.apiCall)}()"
```
It has to look like this instead:
```
      case t: IntType =>
        s"$io.Read${Utils.capitalize(t.apiCall)}()"
      case t: FloatType =>
        s"$io.Read${Utils.capitalize(t.apiCall)}()"
```

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/exprlang/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/exprlang/DataType.scala
@@ -39,6 +39,15 @@ object DataType {
     }
   }
 
+  abstract class FloatType extends BaseType {
+    def apiCall: String
+  }
+  case class FloatMultiType(width: IntWidth, endian: Endianness) extends FloatType {
+    override def apiCall: String = {
+      s"f${width.width}${endian.toString}"
+    }
+  }
+
   trait Processing {
     def process: Option[ProcessExpr]
   }
@@ -85,6 +94,7 @@ object DataType {
   case class EnumType(name: String, basedOn: IntType) extends BaseType
 
   private val ReIntType = """([us])(2|4|8)(le|be)?""".r
+  private val ReFloatType = """f(4|8)(le|be)?""".r
 
   def fromYaml(
     dt: String,
@@ -121,6 +131,22 @@ object DataType {
               defaultEndian match {
                 case Some(e) => e
                 case None => throw new RuntimeException(s"unable to use integer type '${dt}' without default endianness")
+              }
+          }
+        )
+      case ReFloatType(widthStr, endianStr) =>
+        FloatMultiType(
+          widthStr match {
+            case "4" => Width4
+            case "8" => Width8
+          },
+          endianStr match {
+            case "le" => LittleEndian
+            case "be" => BigEndian
+            case null =>
+              defaultEndian match {
+                case Some(e) => e
+                case None => throw new RuntimeException(s"unable to use floating point type '${dt}' without default endianness")
               }
           }
         )


GreyCat:
Yeah, that would be probably a good idea to make an hierarchy or at least a common trait (that's Scala's term for "interface") here. Hang on, I'll try to do it.

GreyCat:
Oh, here it is: https://github.com/kaitai-io/kaitai_struct_compiler/commit/ab2f6db32dfc9f9d927db50ececa48ab57106af0

Sorry, it took me a little longer than I've expected.

The general idea is that we just declare `ReadableType` trait, which is just the equivalent of interface you've proposed. If we don't really care what the internal type is (i.e. when we just want to do that API call), we do just that:

```scala
      case t: ReadableType =>
        s"$io.Read${Utils.capitalize(t.apiCall)}()"
```

... and when we actually care what's the type inside (for example, for type derivation in calculations), we still have the difference between `IntType` and `FloatType`.

BTW, I've also introduced `CalcFloatType` akin to `CalcIntType`, and probably we'll have to introduce a few rules for expression translation (like "multiplying int by float results in float" or something like that).

LogicAndTrick:
Looking good, that should mean that all the compilers will support basic reads of floating points now. I'll add some float expressions into a test case later.

I'm not sure if I'm confident enough to do the expression type rules, I've looked around the expression code and I can see places that probably need to change, but I wouldn't know the best approach with Scala.

GreyCat:
To be honest, all that "expressions" stuff is extremely very undertested now and I haven't yet come up with a efficient way to do the tests.

So far, there are [native Scala tests](https://github.com/kaitai-io/kaitai_struct_compiler/tree/master/jvm/src/test/scala/io/kaitai/struct/translators) for some of the `Translator` stuff, and they do help, but:

* so far they're totally written by hand
* you have to duplicate lots of stuff for all the languages
* what's I'm really aggravated with, is that they tend to horribly go out of sync with each other: i.e. you add tests for `JavaTranslatorSpec`, but you don't add tests for `CSharpTranslator`, `RubyTranslator`, etc, then you forget to do so, and then, kaboom, everything is a mess

Probably we should come up with a way to get these expressions / translator tests to show up at [CI run aggregation](http://kaitai.io/ci) too, but I'm still kind of at loss on how to do that. Plus, even if I'll start enforcing missing tests as failures, it's a huge technical debt to reimplement all translator tests for all languages.

I'm still not sure if current way is a good way to go. Maybe it's worth to rewrite them all in some declarative way, i.e. loading expressions + expected test results from an external file.

LogicAndTrick:
Using an external file would probably work well for those tests since they're mostly comparing strings. Each language result would still need to be written by hand, but having them all together in a flat file would make it faster and easier to add new tests. 

Maybe the translator test results could go in a separate table in the CI aggregation page, because they're tests of the compiler inputs, rather than the compiler outputs.

GreyCat:
I'll probably give it a try, if I'll get to understand how data-driven testing in Scalatest works...

GreyCat:
I've pushed an idea of [saner `Translator` test framework](https://github.com/kaitai-io/kaitai_struct_compiler/commit/102c83fc1c6ef49724e417b03834b967f6b4e5cf), based on scalatest's `Table`.

To make a long story short, now everything boils down to filling `TranslatorSpec.tests` table, which can be filled using a few helper functions (`everybody`, `everybodyExcept`, `full`). One might ignore the gory details beyond `tests` declaration, they're trivial anyway.

I've ported probably all of the existing `TranslatorSpec` tests, now I'll try to add nice HTML report on that to be shown at our /ci page.

GreyCat:
Woohoo, here it is: http://kaitai.io/ci/ - now we have 2 distinct sections, "Expressions" and "KSY compilation", with separate stats for each.

LogicAndTrick:
Awesome! I'll add in the remaining C# test cases and some floating point ones.

GreyCat:
Feel free to :) Probably we'll also need some .ksy tests for floats - that would reveal missing support in runtimes, then we'll do stuff like `read_f4le` in runtimes.

PR #4, opened 2016-07-25
Translator tests: Add floating point tests

LogicAndTrick:
Floating point expressions aren't supported yet so they just throw an exception in the test runner before any results are generated. I wrapped the expression parse in a `Try` type, is that the best way to do it in Scala? I put the exception handler inside the compiler loop so that a failure is generated for each language.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
@@ -7,6 +7,8 @@ import org.scalatest.FunSuite
 import org.scalatest.Matchers._
 import org.scalatest.prop.TableDrivenPropertyChecks
 
+import scala.util._
+
 class TranslatorSpec extends FunSuite with TableDrivenPropertyChecks {
   val tests = Table(
     ("src", "srcType", "expType", "expOut"),
@@ -17,6 +19,11 @@ class TranslatorSpec extends FunSuite with TableDrivenPropertyChecks {
     everybody("1234", "1234"),
     everybody("-456", "-456"),
     everybody("0x1234", "4660"),
+
+    // Float literals
+    everybody("1.0", "1.0", CalcFloatType),
+    everybody("123.456", "123.456", CalcFloatType),
+    everybody("-123.456", "-123.456", CalcFloatType),
 
     // Simple integer operations
     everybody("1 + 2", "(1 + 2)"),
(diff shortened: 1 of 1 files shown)


LogicAndTrick:
I was able to get the parser to work without throwing an exception by switching the order of the float and int expressions in `Expressions.scala`:

```
  val atom: P[Ast.expr] = {
      ... 
      NAME.map(Ast.expr.Name(_)) |
      FLOAT_NUMBER.map(Ast.expr.FloatNum) |
      INT_NUMBER.map(Ast.expr.IntNum)
```

I'll leave this open because I think it's a good idea to have error handling on the expression parser to avoid exceptions while trying to run tests.

LogicAndTrick:
I would have thought the same thing, but expression is parsed outside the `test` block so it's not in the test runner at that point. But now that you mention it, the better option would be to just move the expression parse inside the loop like this:

```
  for ((src, tp, expType, expOut) <- tests) {
    LanguageCompilerStatic.NAME_TO_CLASS.foreach { case (langName, langObj) =>
      test(s"$langName:$src") {

        val e = Expressions.parse(src)  // <-- here

        val tr: BaseTranslator = langObj.getTranslator(tp)
        expOut.get(langObj) match {
          case Some(expResult) =>
            tr.detectType(e) should be(expType)
            tr.translate(e) should be(expResult)
          case None =>
            fail("no expected result")
        }
      }
    }
  }
```

The expression gets parsed a bunch of times, but performance isn't an issue in unit tests...

LogicAndTrick:
That works, could you do that change when you have time? I've not figured out how to work with variable scoping in Scala yet... I'll cancel this PR and re-add the tests afterwards.

GreyCat:
Ok, hang on a sec.

GreyCat:
Here you go: https://github.com/kaitai-io/kaitai_struct_compiler/commit/8e86cfa4fbb616f036027d99e6333b9af17f9b61

LogicAndTrick:
Thanks. I'll add the rest of the tests straight in but I'll create a PR soon for some expression/translator changes that I'm looking at to get some of the floating point tests passing.

GreyCat:
Yeah, feel free to fix the expression AST parser too - that's exactly what I was saying about that it *is* undertested and things like that are bound to come up. Originally, it had no notion of "integers" and "floats" per se, it just had single `Num(Any)` class for all the numbers, so I've added distinct ints and floats recently when I was working on byte arrays. Looks like I've messed it up :(

PR #9, opened 2016-08-11
Unary operations on Int8/Int16 should be cast to Int32

LogicAndTrick:
In C# and Java (and probably other languages), negating a `byte` or `short` value automatically converts it to an `int`. This change emulates that logic. (Not sure if this is good code, it doesn't seem very 'Scala-like', please review)

---

KSY to replicate issue:
```yaml
meta:
  id: cast_to_int
  endian: le
seq:
  - id: value
    type: s2
instances:
  invalid_cast:
    value: -value
```
Compiled result (before change):
```cs
// Error CS0266 - Cannot implicitly convert type 'int' to 'short'. An explicit conversion exists (are you missing a cast?)
class CastToInt : KaitaiStruct
{
    ...
    
    private short _invalidCast;
    public short InvalidCast
    {
        get
        {
            ...
            _invalidCast = -Value;
            ...
        }
    }
}
```
After change:
```cs
// No compile errors
class CastToInt : KaitaiStruct
{
    ...
    
    private int _invalidCast;
    public int InvalidCast
    {
        get
        {
            ...
            _invalidCast = -Value;
            ...
        }
    }
}
```

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
@@ -196,8 +196,10 @@ abstract class BaseTranslator(val provider: TypeProvider) {
       case Ast.expr.UnaryOp(op: Ast.unaryop, v: Ast.expr) =>
         val t = detectType(v)
         (t, op) match {
-          case (Int1Type(false), _) => CalcIntType
-          case (_: IntType, _) => t
+          case (intMultiType: IntMultiType, _) =>
+            if (intMultiType.width.width > 4) t
+            else CalcIntType
+          case (_: IntType, _) => CalcIntType
           case (_: FloatType, Ast.unaryop.Minus) => t
           case _ => throw new RuntimeException(s"unable to apply unary operator ${op} to ${t}")
         }


GreyCat:
This one is a hard decision to make, but probably we should just roll this way. While Java and C# have this upcasting logic, other languages may do or do not implement this same. This one is a compile error in Scala (as in Java too):

```scala
val v1: Byte = 42
val v2: Byte = -v1 // -v1 has derived type "Int"
```

OTOH, this works perfectly:

```scala
val v1: Byte = -42
```

Probably because `-42` is not parsed as `42` + unary minus operator, but as a regular negative integer literal.

And this one works perfectly ok in C/C++:

```c
int8_t v1 = 42;
int8_t v2 = -v1;
```

No warning, no problems.

LogicAndTrick:
Maybe this logic should be moved into the language-specific translators?

GreyCat:
That would create an even more impressive sort of mess: if we'll allow `detectType` to derive different types based on the language... We could then get .ksys that are compilable to *some* target languages, but fail compiling on the others...

GreyCat:
I don't know if you really want to dig deeper into Scala or not, but in case you'll be interested - that's a little bit more Scala-esque version of your code - https://github.com/kaitai-io/kaitai_struct_compiler/commit/006bee81fac1a5c626d8b6a533c294db066b4a46 - this one uses matching + if guards.

LogicAndTrick:
Ah, so that's how it's done! Thanks for letting me know. This Scala stuff has been a fun learning experience for me so far, previously I've only worked with traditional OO languages :)

Am I right in thinking that line 200 is just there for code clarity and could be removed? It would fall through to the next line which just returns `CalcIntType` for any `IntType`.

GreyCat:
Come to think of it, you're probably right :) It's not "for code clarity", it's just me blindly overlooking such stuff. I think we really need to add your `cast_to_int.ksy` as a test.

LogicAndTrick:
Agreed, when I get some time I'll expand it to include some values for all the integer types so the behaviour for long values can be tested as well. Unsigned long types can't be negated in C#/Java so I'll leave that out. 

I suspect that unsigned ints might cast to long when negated which would require another change. There's also a whole other range of issues related to binary operations with longs, which will all fail because `CalcIntType` is a 32-bit integer. Should I add those into the same test too?

Example:
```yaml
meta:
  id: cast_to_long
  endian: le
seq:
  - id: value
    type: s8
instances:
  calcinttype_is_not_long:
    value: value + 1    # Error CS0266: Cannot implicitly convert type 'long' to 'int'.
```

GreyCat:
I've added it as [type_int_unary_op.ksy](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/formats/type_int_unary_op.ksy) with 2 test cases. Feel free to exapnd.

PR #11, opened 2016-08-12
PythonCompiler: remove from_file() helper

antonv6:
This will make the generated code smaller.

(Obviously needs the method to be present in KaitaiStruct instead)


GreyCat:
Thanks! Seems to work, soon it shall go through Travis and results will be at http://kaitai.io/ci

antonv6:
TypeIntUnaryOp on Python says "didn't run", but other tests are all green. I assume it's fine then?

GreyCat:
Yeah, that's a new test, so no specs were done for all the languages yet. We'll add them eventually.

PR #13, opened 2016-08-16
Formal names

GreyCat:
Well, this looks like a *big* commit, but in reality, it's all about massive refactoring, no new functionality here.

The basic idea is simple: previously, we've worked with `String` as identifiers, using `String` both for identifier from the .ksy file itself (i.e. `foo_bar`), rendition of it to language-specific style `fooBar` and rendition of it with some special stuff appended/prendended to be invoked as a private member name or something (i.e. `this.fooBar` or `self.foo_bar` or `@foo_bar` or something).

It created a huge mess with all that chain of `lowerCamelCase` functions that were abused and that lead to lots of code duplication and hardships while developing new languages support. For example, it would be great to just change `this.` into `$this->` in exactly one place and kind of convert Java into PHP.

So, here's what I came up with:

* All identifiers are now not the simple `String`, but a string wrapped in some subclass of `Identifier`.
* It is illegal to use identifier as is, embedding / concatenating it to the string. Before usage, identifier must undergo some transformation into a string, there are now 3 of them:
  * `idToStr` - basic transformation used in a language (`foo_bar` -> `fooBar`)
  * `privateMemberName` - transform identifier to something that's ready to be used as a reference to private member (`foo_bar` -> `@foo_bar` in Ruby, `this.fooBar` in Java, `_fooBar` in C# or something along the lines)
  * `publicMemberName` - transform identifier to the name of public member (i.e. a getter method, or property, or something like that — `foo_bar` -> `foo_bar` in Ruby, `fooBar` in Java, `FooBar` in C#).

There are no more plain string mangling, appending cryptic `_io_`, `_raw_`, etc, prefixes. If you want an additional identifier for a raw byte array for a given identifier, it is `RawIdentifier(existingId)` - voila, that's neatly wrapper, and all id-to-string functions are expected to work properly with that. The same stuff mostly applies to temporary IO vars.

In future, probably it should be very easy to add options to override default languages identifier transformation rules using this thing.

It's still not a very polished code (but it seems to pass the tests), so I would really love to hear some thoughts and feedback on it, especially from @LogicAndTrick, who seems to be working on support for yet another language.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
@@ -167,36 +167,34 @@ class ClassCompiler(val topClass: ClassSpec, val lang: LanguageCompiler) extends
     lang.instanceFooter
   }
 
-  override def determineType(attrName: Identifier): BaseType = determineType(nowClass, nowClassName, attrName)
+  override def determineType(attrName: String): BaseType = determineType(nowClass, nowClassName, attrName)
 
-  override def determineType(typeName: List[String], attrName: Identifier): BaseType = {
+  override def determineType(typeName: List[String], attrName: String): BaseType = {
     getTypeByName(nowClass, typeName) match {
       case Some(t) => determineType(t, typeName, attrName)
       case None => throw new RuntimeException(s"Unable to determine type for $attrName in type $typeName")
     }
   }
 
-  def determineType(classSpec: ClassSpec, className: List[String], attrName: Identifier): BaseType = {
+  def determineType(classSpec: ClassSpec, className: List[String], attrName: String): BaseType = {
     attrName match {
-      case RootIdentifier =>
+      case "_root" =>
         UserTypeInstream(topClassName)
-      case ParentIdentifier =>
+      case "_parent" =>
         UserTypeInstream(classSpec.parentTypeName)
-      case IoIdentifier =>
+      case "_io" =>
         KaitaiStreamType
-      case namedId: NamedIdentifier =>
+      case _ =>
         classSpec.seq.foreach { el =>
-          if (el.id == attrName)
+          if (el.id == NamedIdentifier(attrName))
             return el.dataTypeComposite
         }
-        throw new RuntimeException(s"Unable to access ${namedId.name} in $className context")
-      case instId: InstanceIdentifier =>
-        classSpec.instances.get(instId) match {
+        classSpec.instances.get(InstanceIdentifier(attrName)) match {
           case Some(i: ValueInstanceSpec) => return i.dataType.get
           case Some(i: ParseInstanceSpec) => return i.dataTypeComposite
           case None => // do nothing
         }
-        throw new RuntimeException(s"Unable to access ${instId.name} in $className context")
+        throw new RuntimeException(s"Unable to access $attrName in $className context")
     }
   }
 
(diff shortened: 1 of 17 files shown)


LogicAndTrick:
I like this change, it will remove a lot of the confusion when implementing a new language compiler. 
A few ideas/questions:

- Are the `[public|private]MemberName` functions in the compilers related at all to the `doName` method in the translator? In some of the translators, there is duplicated logic for identifiers in the `doName` method. Can the two `MemberName` methods be moved into the translators, or can this logic be merged in some other way?

- Public and private member names are not the only possible reference names - for example, there might be a `localVariableName` which might return `identifier` instead of `this.identifier`. This issue is already showing a bit in the `allocateIO` method of some compilers. How about some way to abstract away the different types of identifiers? Rough example:

```scala
abstract class TypedIdentifier
case class PublicMember(identifier: Identifier) extends TypedIdentifier
case class PrivateMember(identifier: Identifier) extends TypedIdentifier
case class LocalVariable(identifier: Identifier) extends TypedIdentifier
// etc...

// Compiler or maybe Translator
override def identifierName(type: TypedIdentifier) = type match {
    case PublicMember(id) => ...
    case PrivateMember(id) => ...
    case LocalVariable(id) => ...
    // etc...
}

// Compiler
override def handleAssignmentRepeatEos(id: Identifier, expr: String): Unit = {
    out.puts(s"${identifierName(PublicMember(id))}.add($expr);")
}
override def allocateIO(varName: Identifier, rep: RepeatSpec): String = {
    val ioName = identifierName(LocalVariable(varName));
    ...
}
```

GreyCat:
> Are the `[public|private]MemberName` functions in the compilers related at all to the doName method in the translator?

It's complicated - both "yes" and "no", mostly "no". At first, I tried to introduce `Identifier` into translators too, but soon enough I realized that they're kind of 2 relatively independent projects merged into one we have here. The thing is, when you write `foo_bar` or `_io` or `_raw_foo` in translators' expression, there is no way to determine whatever it is - is it a reference to the private member, public member, a local variable, is it sequence attribute or instance reference. Even `TypeProvider` does not actually provide that information.

I guess, for a baseline, we can assume that translators' expressions never access private members of the classes - they always work through public interfaces, i.e. getters. And `publicMemberName` is actually about defining public members, not calling them, i.e. `publicMemberName` doesn't do stuff like `->` or `()`, only the name itself.

Actually, we might do a simple test: trying changing public member naming scheme temporarily to something else and see what gets broken.

> How about some way to abstract away the different types of identifiers?

I like the idea, but it adds an extra layer of complexity (and probably compilers nowadays are even more complicated to read without investing major time into understanding what's going on, as you did ;)). Having distinction between `Identifier` and plain `String` is kind of important, as it allows a clearer separation of "business logic" (i.e. wrapping of identifier with a RawIdentifier to generate new "raw" type attribute, related to existing attribute) and "presentation" (i.e. actual rendering of id to string in one of the possible ways). And I see no point in *storing* the state of public/private/local — except, may be, for local  variables — for replacement of that pesky `io: String` stuff.

Then again, there's also huge untouched area of class/type names, its hierarchy, etc, which is currently relayed over as `List[String]`, and an area of enums. These are candidates for further improvement too.

GreyCat:
That said, I believe we've both ok with this change. So I'm merging it to master.

LogicAndTrick:
Yep, no complaints here. I think the `doName` method in the translator is a potential point of confusion, could it be renamed to something more obvious? Or maybe just a scaladoc comment on the `BaseTranslator` method would be enough.

GreyCat:
I start to feel that *lots* of naming in this project might need a review. `ClassCompiler` was a bad choice. `LanguageCompiler` should be probably replaced with something more obvious. `BaseType` needs a better name (as it is really "any valid KS data type"). That whole "translators" vs "compiler" terminology is a mess, as, technically "compiler" is also a "translator". Many method names might benefit from a rename too. And, yeah, adding more scaladoc to base classes definitely won't hurt.

PR #16, opened 2016-08-31
Switch

GreyCat:
Implemented very basic type switching: format parsing + language support for Ruby, Java, C# and Python.

The syntax is:

```yaml
  - id: code
    type: u1
  - id: body
    type:
      switch-on: code
      cases:
        73: intval
        83: strval
```

or, with enums (implies that `code_enum` is defined):

```yaml
  - id: code
    type: u1
    enum: code_enum
  - id: body
    type:
      switch-on: code
      cases:
        code_enum::intval: intval
        code_enum::strval: strval
```

No `else` case support so far, support for anything else besides user types is not tested (should work for duck-typed languages, would fail for C# and Java).

There are 2 new tests: `switch_manual_int` and `switch_manual_enum` already committed to tests repo. C#, Java and Ruby implementations work (although C# again has name clashes of enums vs attributes). Python implementation probably works too, but right now it's held back by class addressing issues.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/exprlang/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/exprlang/DataType.scala
@@ -108,6 +108,8 @@ object DataType {
 
   case class EnumType(name: String, basedOn: IntType) extends BaseType
 
+  case class SwitchType(on: Ast.expr, cases: Map[Ast.expr, BaseType]) extends BaseType
+
   private val ReIntType = """([us])(2|4|8)(le|be)?""".r
   private val ReFloatType = """f(4|8)(le|be)?""".r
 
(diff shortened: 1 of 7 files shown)


GreyCat:
Ok, I guess, no objections, merging it to master then.

LogicAndTrick:
Sorry, I haven't had time to try this out yet. Has something like this been tested (repeat + switch at the same time)? Does it choose the correct base type for the array in the strongly typed languages?

```yaml
  - id: code
    type: u1
  - id: body
    type:    # <-- switch
      switch-on: code
      cases:
        73: intval
        83: strval
        99: u1     # <-- what happens when a primitive type is one of the options?
    repeat: expr      # <-- repeat
    repeat-expr: 10
```

GreyCat:
There is a lot to test and do here. So far, I haven't made proper type derivation for strongly typed languages, there's just a plug is most of the cases there ;)

```scala
  def kaitaiType2JavaTypePrim(attrType: BaseType): String = {
    attrType match {
      // ...
      // TODO: derive type properly, we might actually use some primitive type there or something
      case SwitchType(on, cases) => kstructName
    }
  }
```

Not only that, but actually we have to test stuff like having `size: XXX` set up, but not all used types support it. Etc, etc.

PR #18, opened 2016-10-07
C# Compiler: Force cast calculated instance values 

LogicAndTrick:
Allows the ExprMod test to be compiled.

Note that the test still won't actually *pass* at this point because of the way `%` on negative numbers differs between languages. But this change is required so it will actually compile.

I'm not sure that this project should enforce mod consistency between languages to begin with, but that's another discussion entirely. On that note, [Wikipedia has a ridiculously comprehensive list](https://en.wikipedia.org/wiki/Modulo_operation#Remainder_calculation_for_the_modulo_operation) of how different languages behave when modding negative numbers.


GreyCat:
About mod consistency - yeah, I'm pretty much sure that it's the right way. It's one of the most frequently used operations in offset / indexes calculation — and it would be very inconvenient to have it behaving differently in different target languages.

LogicAndTrick:
Fair enough, that makes sense. And the "always positive" result is definitely the more useful result for offset calculations, so the tests are right.

I don't know if there's a nice way to calculate it... maybe something like:
```
a % b = ((a % b) + b) % b;
```
Looks ugly but it'd do the job I guess.

PR #20, opened 2016-10-18
Add CDATA to C# XML docstrings

LogicAndTrick:
XML special characters have to be escaped in C# docstrings, but this can be avoided by using a CDATA section. There's no harm in putting everything inside CDATA, so this is easier than performing an XML escape.

(Technically if the docstring contains the characters `]]>` then it'll still break, so maybe an XML escape is a better idea?)


GreyCat:
One of our goals is to keep generated files readable, so we should probably avoid adding escapes where it's not needed. I don't know which escaping should we prefered from readability POV, though: XML or CDATA...

LogicAndTrick:
XML escape is probably the best solution in the long term because it'll cover all situations, but I don't know the best approach in Java/Scala (does it have any XML libraries included that will do a proper escape?).

The quick and dirty fix would be to do a string replace for `<`, `>`, and `&` with their escaped variants. I don't think any other (commonly used) character would need escaping for C# docstrings. CDATA will do the job for now though, I doubt `]]>` would appear in many docstrings anyway.

PR #26, opened 2016-10-31
JavaScriptCompiler: proper partial object parse, improved code readability

koczkatamas:


shared/src/main/scala/io/kaitai/struct/languages/JavaScriptCompiler.scala
```diff
@@ -64,9 +64,16 @@ class JavaScriptCompiler(config: RuntimeConfig, out: LanguageOutputWriter)
     out.puts("this._io = _io;")
     out.puts("this._parent = _parent;")
     out.puts("this._root = _root || this;")
-    if(debug)
+    if(debug){
```
GreyCat: Ok, keep trying ;) You'll get that eventually - `if (debug) {` :+1: 
Please take a look at http://docs.scala-lang.org/style/control-structures

PR #27, opened 2016-11-02
Unified YAML parsing

GreyCat:
This is a somewhat complete re-implementation that addresses issue https://github.com/kaitai-io/kaitai_struct/issues/42. All internal structures are re-implemented to accommodate raw SnakeYAML + objects coming from JavaScript YAML parser. Jackson mapping library that we've used previously, is removed.

As far as I see, all tests pass, new compiler generates the same code as old one. As a bonus, we've got much nicer error messages.


GreyCat:
Looks like everything works, everybody's happy :)

Issue #29, opened 2016-11-21
CTOR mismatch in Java in case of types in different files.


tschoening:
I want to use types in different files, pretty much like you suggest in [some other issue comment](https://github.com/kaitai-io/kaitai_struct/issues/48#issuecomment-261751957) and are using for the [network examples](https://github.com/kaitai-io/kaitai_struct_formats/tree/master/network). The problem I have is that my generated code doesn't compile in Java and that's the same with your network examples: I was compiling ethernet_frame with its dependencies and have the same result, that the generated code is invalid.

Have a look at some CTORs:

    public EthernetFrame(KaitaiStream _io, KaitaiStruct _parent, EthernetFrame _root) throws IOException {
        super(_io);
        this._parent = _parent;
        this._root = _root;
        _read();
    }

    public Ipv4Packet(KaitaiStream _io, KaitaiStruct _parent, Ipv4Packet _root) throws IOException {
        super(_io);
        this._parent = _parent;
        this._root = _root;
        _read();
    }

And how those are called:

    this.ipv4Body = new Ipv4Packet(_io__raw_ipv4Body, this, _root);

All CTORs including a `_root` seem to have themselfs as type for the root, which make sense if the ksy files are self contained, but seems to break in a parent/child relation ship. In such a case the root would need to be either a constant root type or it would need to be same type like parent or maybe some constant/parent-like root and additionally some child specific root?

Or am I doing something wrong?

I'm using the current master of the compiler.

GreyCat:
This is indeed a bug, thanks for reporting that. As far as I can see, one obvious solution is to stop supplying  our `_root` as `_root` (and, of course, parent as well) for all the classes that are not declared in current file.

tschoening:
With that approach one couldn't access the entire `EthernetFrame` from within `Ipv4Packet` anymore, right?

Don't you know already about the parent-child-relationship in `Ipv4Packet` and could simply add one CTOR with a root of type `EthernetFrame`? In effect you would introduce two roots, one local to the file with the type of the file and one global belonging to the parent of the file, if any. Something like the following:

    public Ipv4Packet(KaitaiStream _io, KaitaiStruct _parent, Ipv4Packet _self|localRoot) throws IOException {


    public Ipv4Packet(KaitaiStream _io, KaitaiStruct _parent, EthernetFrame _[parent|global]root) throws IOException {

[...]

PR #45, opened 2016-12-12
#44: #29 ch…

tschoening:
…anged the CTORs called in case of opaque types and it looks this change was missed for Ruby by accident. With this change #44 is fixed.


GreyCat:
Thanks for the report and the solution! If I could ask of anything more, could you provide a test for this, so we'll see that it was broken and it got fixed afterwards?

tschoening:
I've added pull request [17](https://github.com/kaitai-io/kaitai_struct_tests/pull/17) in the tests repo with one test for Ruby, as this is what fails and I don't have all the other runtimes anyway.

GreyCat:
Ruby-only spec is usually more than enough: all others can be generated automatically anyway using `spec_translate_missing`.

tschoening:
Do you plan on pulling this or is there anything still missing I should look into?

GreyCat:
Everything's fine, thanks a lot!
Sorry for the delays, I'm kind of short of time due to all that end-of-year chaos...

Issue #46, opened 2016-12-14
Missing array-creation for debug repeat-until


Glitch0011:
Hi et al

I'm finding that the web-IDE is showing me an error whilst using repeat-until to parse a DNS dissector I'm writing.

![capture](https://cloud.githubusercontent.com/assets/6094389/21184527/c39d2944-c204-11e6-8111-010f9c208a15.png)

I've looked through the code and believe I've found the bug to be on line #256 in JavaScriptCompiler.scala (warning, I don't know scala so intuitive guess work here).

My guess (from the way condRepeatEosHeader handles it) is a simple insert will fix the bug. I've noticed that #25 was the commit related to adding Debug mode.

I have the commit if you'd like me to do it?

Tom

GreyCat:
Thanks for reporting and proposing a solution! I will investigate it soon.

koczkatamas:
@GreyCat updated the compiler and I updated the WebIDE, could you try it again?

Also please check that you have the latest version by looking at the WebIDE about window, it should say "Kaitai compiler version: 0.6-SNAPSHOT (2016-12-14 19:38:13.871)".

Thanks!

Glitch0011:
Can confirm, issue is fixed and correctly parsing DNS's interesting repeat-structure.

![fix](https://cloud.githubusercontent.com/assets/6094389/21217141/943dc3ea-c2a2-11e6-8144-b50970dbe563.PNG)

Thanks

GreyCat:
Thanks! Could you consider contributing that DNS-related protocol you've described to our [format repo](http://github.com/kaitai-io/kaitai_struct_formats)?

PR #47, opened 2016-12-14
#46 Fix to non-initialised array

Glitch0011:
https://github.com/kaitai-io/kaitai_struct_compiler/issues/46 Added debug array creation to repeat-until header to prevent TypeError exception


Glitch0011:
I will warn that I don't know scala and cannot test currently, so please review. However it seems pretty logical so that commit should work :)

GreyCat:
Looks good, thanks! My only concert is broken formatting, I'll fix that.

Issue #38, opened 2016-12-06
Parent keyword for switch construction


zeddvandaar:
Parent::child construction does not working for "switch-on" expression.

Real example of blizzard resources file format.

```
meta:
  id: m3
  application: Blizzard Game
  endian: le
seq:
  - id: header
    contents: ["43DM"]
  - id: ofs_tags
    type: u4
  - id: n_tags
    type: u4
types:
  tag:
    seq:
      - id: name
        type: str
        size: 4
        encoding: ASCII
      - id: offset
        type: u4
      - id: nbr_items
        type: u4
      - id: flag
        type: u4
    types:
      # tag_sd:
        # types:
          # tag_sd_item:
            # seq:
              # - id: entries
                # type: u4
              # - id: indx
                # type: u4
              # - id: flag
                # type: u4
              # - id: flags
                # type: u4
              # - id: end_frame
                # type: u4
              # - id: entries2
                # type: u4
              # - id: indx2
                # type: u4
              # - id: flag2
                # type: u4
        # instances:
          # tag_sd_content:
            # type: tag_sd_item
            # repeat: expr
            # repeat-expr: nbr_items
      tag_char:
        seq:
          - id: content
            # type: u1
            type: str
            size: _parent.nbr_items
            encoding: ASCII
    instances:
      tag_content:
        pos: offset
        type: 
          switch-on: name
          cases:
            # '"VEDS"': tag_sd
            # '"V2DS"': tag_sd
            # '"SD3V"': tag_sd
            # '"3RDS"': tag_sd
            # '"CCDS"': tag_sd
            # '"6SDS"': tag_sd
            # '"6UDS"': tag_sd
            # '"3UDS"': tag_sd
            # '"Q4DS"': tag_sd
            # '"GFDS"': tag_sd
            # '"BMDS"': tag_sd
            '"RAHC"': tag_char
        # repeat: expr
        # repeat-expr: nbr_items
        io: _root._io
instances:
  index:
    pos: ofs_tags
    type: tag
    repeat: expr
    repeat-expr: n_tags
    io: _root._io
```

nbr_items field is avaidle at tag_content instance, but _parent.nbr_items does not working at included over switch-on element.
I propose to change behavior of _parent expression, or add _owner keyword for runtime linking.

GreyCat:
Thanks for reporting this. There was a similar bug like that: https://github.com/kaitai-io/kaitai_struct/issues/48 — but this seems to be yet another untested case.

[...]

PR #49, opened 2016-12-16
#38: fix _parent not derived in value instances

tschoening:
This is a first attempt to fix the problem that `parent` can't be resolved if it's used in `value` instances, because the former would normally be resolved after the latter. I couldn't find an easy way to switch the processing order of both, because during resolving `_parent` one needs "something" to assign the parent to and I don't see what this should be.

So I tried by simply collection parent-child-relationship on its own and provide that only for processing fo `value` instances and that worked. My KSY compiles now and the tests still pass as well. Not sure if this si the right approach, though.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
@@ -5,7 +5,7 @@ import io.kaitai.struct.format._
 import io.kaitai.struct.translators.{TypeMismatchError, TypeProvider, TypeUndecidedError}
 
 class ClassTypeProvider(topClass: ClassSpec) extends TypeProvider {
-  var nowClass           : ClassSpec         = topClass
+  var nowClass : ClassSpec = topClass
   var possibleParentClass: Option[ClassSpec] = None
 
   var _currentIteratorType: Option[BaseType] = None
--- a/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
@@ -5,7 +5,7 @@ import io.kaitai.struct.format._
 import io.kaitai.struct.translators.{TypeMismatchError, TypeProvider, TypeUndecidedError}
 
 class ClassTypeProvider(topClass: ClassSpec) extends TypeProvider {
-  var nowClass : ClassSpec = topClass
+  var nowClass: ClassSpec = topClass
   var possibleParentClass: Option[ClassSpec] = None
 
   var _currentIteratorType: Option[BaseType] = None
--- a/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
@@ -5,7 +5,8 @@ import io.kaitai.struct.format._
 import io.kaitai.struct.translators.{TypeMismatchError, TypeProvider, TypeUndecidedError}
 
 class ClassTypeProvider(topClass: ClassSpec) extends TypeProvider {
-  var nowClass = topClass
+  var nowClass           : ClassSpec         = topClass
+  var possibleParentClass: Option[ClassSpec] = None
 
   var _currentIteratorType: Option[BaseType] = None
   var _currentSwitchType: Option[BaseType] = None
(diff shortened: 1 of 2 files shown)


tschoening:
Did you have a chance to review this pull request and the corresponding test yet? Is there something I need to change? I'm not sure about what the bot says, I'm no Scala expert and don't see any real issue with the current implementation or good alternatives worth changing things.

I would like to not leave this request orphaned, because I need the changes for my KSY to work properly and would like to stick as close to your master branch as possible in my forks.

Thanks!

GreyCat:
Well, better late than never, they say ;)

I've took a closer look and there are some news.

The bad news is that I'm sorry, but I can't accept your pull request: it more like masks the problem by building an alternative name resolution mechanism (that would likely fail with more intricate name resolution that won't fit into `parentsByChild` map scheme) — and, besides, it relies heavily on internal state (`var`) on `ClassTypeProvider`, and introduces more conditions on call order of the procedures there.

The good news is that I've submitted another solution: actually, it's quite simple, it just swaps `markupParentTypes(topClass)` and `deriveValueTypes(topClass)` around, but with a twist. Previously, for "parenting" purposes, all instances (including *both* ParseInstances and ValueInstances) were considered. In fact, that's a misconception: value instances should never be used for purposes of determining a parent. For example, this was kind of broken before:

```yaml
meta:
  id: main_type
seq:
  - id: some_root_attr
    type: u1
  - id: child
    type: child
  - id: child2
    type: child2
types:
  child:
    seq:
      - id: foo
        size: _parent.some_root_attr
  child2:
    instances:
      root_child:
        value: _root.child
```

In this example, `child2` has reference to `root_child` in a value instance. Older algorithm would have considered that both `main_type` and `child2` are parents for `child`, thus creating a collision and making it generic `KaitaiStruct` type. Newer algorithm ignores value instances for parenting purposes and thus you'll get `_parent` to be `main_type`, and thus `_parent.some_root_attr` would be resolved properly.

In the process, I've added some debugging output with more-or-less proper logging "framework". These are now available to see using `--verbose all`, `--verbose parent`, etc. One can select which subsystem(s) to enable logging for.

tschoening:
My KSY still works, so thanks for your solution. It's a bit annoying that I didn't find it myself, though. :-) I considered mine as hack as well, but it worked...

GreyCat:
Actually, diving into this brought up quite a few subjects that need(ed) some attention:

* Proper logging system inside the application, especially to track non-trivial things like type derivation mechanisms in [TypeProcessor](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/shared/src/main/scala/io/kaitai/struct/TypeProcessor.scala) — kind of implemented now, but probably still needs tons of work to make it more legible and useful.
* Lack of documentation in complex things like TypeProcessor; after having a 1-2 month break and revisiting it, it took me some time to get to understand what's going on there. Still needs to be addressed.
* What we want to consider "parenting" of an object (and thus what determines parent type). The trivial solution (more or less what we do now) is "every time when we do `new Child`, we pass `this` to it as `parent` → so, that's the parent type". However, even that is not always the best solution. For example, sometimes would rather supply something else besides `this`, for example, `this._parent` as a parent, or would rather prefer to instantiate the type without any parent at all.

tschoening:
> For example, sometimes would rather supply something else besides this, for example, this._parent as a
> parent[...]

Good idea and I have exactly that use-case. :-) The following are the important parts of my KSY as an example:

    fmt_oms_apl_payload:
      seq:
        - id:   vdbs_decrypted_wo_enc_mode
          type: fmt_oms_apl_vdb
          repeat:       until
          repeat-until: _.next_isnt_vdb
          if:   not is_error_instead_of_vdbs and are_vdbs_decrypted_wo_enc_mode
        - id:   vdbs_decrypted_with_enc_mode
          type: fmt_oms_apl_vdbs_decrypted_with_enc_mode
          if:   not is_error_instead_of_vdbs and are_vdbs_decrypted_with_enc_mode
        - id:   vdbs_encrypted
          size: header_short_if.config_field.aes_byte_cnt
          if:   not is_error_instead_of_vdbs and are_vdbs_encrypted

    fmt_oms_apl_vdbs_decrypted_with_enc_mode:
      seq:
        - id:       aes_verify
          contents: [0x2F, 0x2F]

        - id:       var_data_blocks
          type:     fmt_oms_apl_vdb
          repeat:       until
          repeat-until: _.next_isnt_vdb

    fmt_oms_apl_vdb:
      seq:
        - id:   header
          type: fmt_oms_apl_vdb_header

        - id:   data
          size: header.data_size
          type: fmt_oms_apl_vdb_data
          if:   header.data_size > 0

The problem is in *_apl_vdb, which is used by two different parent types, either *_apl_payload directly or *_with_enc_mode. Because of that, it gets CTORs with `KaitaiStruct` as _parent instead of e.g. *_apl_payload and one can't access *_apl_payload using _parent from *_apl_vdb anymore.

        public FmtOmsAplVdb(KaitaiStream _io, KaitaiStruct _parent) {
            super(_io);
            this._parent = _parent;
            _read();
        }

        public FmtOmsAplVdb(KaitaiStream _io, KaitaiStruct _parent, FmtOmsRec _root) {
            super(_io);
            this._parent = _parent;
            this._root = _root;
            _read();
        }

If one could decide somehow to ignore *_with_enc_mode as _parent, that would be great. In my case that type is only used for documentation purpose and the applied checks, it's currently not useful in any other way and doesn't make a good _parent in theory.

PR #50, opened 2016-12-17
JavaScriptCompiler: support for better debug enum information

koczkatamas:


GreyCat:
I did a partial merge — i.e. adding attrType everywhere and extra enum string stuff in JavaScriptCompiler. If we're going to do hex output for enums, let's (1) discuss it, (2) implement it in every compiler.

koczkatamas:
I removed the hex representation changes and fixed the enum name handling, now it uses full type path for enumName (root.innerType.enumType instead of enumType).

PR #51, opened 2016-12-19
#39: Invalid "switch"-statement generated for Java.

tschoening:
In case of multiple boolean operations as condition of a switch statement, parenthesis weren't properly generated. This at least in Java created code which didn't compile.

https://github.com/kaitai-io/kaitai_struct_compiler/issues/39


GreyCat:
Merging this fix for now: in future it will be probably overtaken by https://github.com/kaitai-io/kaitai_struct/issues/69 - but given that I've stumbled upon some problems there and I can't say for sure how much time it will take to progress - this solution is obviously better than no solution.

Sorry it took me so long ;)

Issue #52, opened 2016-12-23
Missing parenthesis around ternary operation in Javascript


koczkatamas:
This
`b1 * (1 == 0 ? 128 : 1)`
generates
`(this.b1 * 1 == 0 ? 128 : 1);`
instead of
`(this.b1 * (1 == 0 ? 128 : 1));`

GreyCat:
Probably it's duplicate of https://github.com/kaitai-io/kaitai_struct_compiler/issues/39 ?

GreyCat:
I've just checked, this particular bug seems to be solved, now it generates:

```javascript
      this._m_foo = (this.b1 * (1 == 0 ? 128 : 1));
```

Thus I'm closing this bug, and probably all other activities on this elusive operator parenthesis / priority question should go into https://github.com/kaitai-io/kaitai_struct/issues/69.

PR #53, opened 2016-12-23
Fix ternary operator translation (add parenthesis around)

koczkatamas:
Fix this issue: https://github.com/kaitai-io/kaitai_struct_compiler/issues/52

But I am not sure that this will solve every possible problems. Also I don't know how to write tests for it.


GreyCat:
Expression translators are generally tested using [TranslatorSpec](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala).

There are currently two fixes: your and https://github.com/kaitai-io/kaitai_struct_compiler/pull/51. We need to investigate and find the minimal amount that will solve the problem without creation of too many extra parenthesis.

GreyCat:
Again, probably the "optimal" solution will come with https://github.com/kaitai-io/kaitai_struct/issues/69, but this solution is better than no solution at all for the time being. Sorry it took me so long.

PR #55, opened 2016-12-30
CppCompiler: fixed switch

adamiwaniuk:
Switch on no Int/Enum type doesn't work properly when there is more than 2 cases (e.q. png.ksv). If length is fixed and switch match some not last case then data is consumed 2 times.


GreyCat:
Thanks for this contribution! Can you also submit a test case into [our tests repo](https://github.com/kaitai-io/kaitai_struct_tests) that illustrates the problem?

adamiwaniuk:
Test case: https://github.com/kaitai-io/kaitai_struct_tests/pull/26

GreyCat:
Thanks again! Off [to CI](http://kaitai.io/ci) it goes, results should be there in 5-7 minutes.

PR #56, opened 2016-12-31
JavaScriptCompiler: add enum name into debug for instances too

koczkatamas:


shared/src/main/scala/io/kaitai/struct/languages/JavaScriptCompiler.scala
```diff
@@ -190,6 +195,10 @@ class JavaScriptCompiler(config: RuntimeConfig, out: LanguageOutputWriter)
     out.puts(s"$debugName.end = $io.pos;")
   }
 
+  override def instanceDebug(instName: InstanceIdentifier, dataType: BaseType, value: Ast.expr): Unit = {
+    attrDebugStart(instName, dataType, null, NoRepeat);
```
GreyCat: Using `null` in our own code is totally not a way to go. If we need `attrDebugStart` to accept either a `io: String` or nothing, it should be declared as `io: Option[String]`, and here we'll supply `None`, and in all other cases we'll give it `Some(...)`.

shared/src/main/scala/io/kaitai/struct/languages/components/EveryReadIsExpression.scala
```diff
@@ -225,6 +225,7 @@ trait EveryReadIsExpression extends LanguageCompiler with ObjectOrientedLanguage
 
   def attrDebugStart(attrName: Identifier, attrType: BaseType, io: String, repeat: RepeatSpec): Unit = {}
   def attrDebugEnd(attrName: Identifier, attrType: BaseType, io: String, repeat: RepeatSpec): Unit = {}
+  def instanceDebug(instName: InstanceIdentifier, dataType: BaseType, value: Ast.expr): Unit = {}
```
GreyCat: That looks like a pretty strange API for me. We have 2 kinds of instances: parse instance and value instance, and this `instanceDebug` lacks `io`, which should be actually used for parse instances. If this is value instance only method, it should be named something like `valueInstanceDebug` or something like that.

GreyCat:
To summarize, what I propose is:

1. Remove `instanceDebug`, it's misleading name
2. Rework `attrDebugStart` to make `io` optional (i.e. `Option[String]`)
3. Add some comments about what it's supposed to do in the first place (i.e. it is needed mostly for JavaScript, because it requires additional info on enum types in debug)
4. Rework unnecessary usage of ListBuffer, try to stay clean & functional; it can be somewhat tricky, not top priority, I can do that after merge.

koczkatamas:
I refactored the code. Could you review it again?

Issue #57, opened 2017-01-04
Try to reduce IOException in instance signatures in Java?


tschoening:
I use a lot of instances in my KSY which actually don't parse anything from the stream, but more or less only work as some getter to map different conditions to some `enum` or such. The point is, that the data fields used in those instances are mostly getters or maybe other instances behaving the same, not parsing anything from the stream as well. So there's no way an `IOException` could be thrown on runtime, hence all signatures of all instances declare exactly that. And because it's a checked exception in Java, every caller of those instances need to handle or throw those unnecessary exceptions as well.

So how about finding ways to reduce this?

Looking at the code, `IOException` is simply added unconditionally in `JavaCompiler.instanceHeader` and currently one doesn't know much about what the instance does. But you already distinguish value and parsing instances in the specs, so wouldn't it be enoguh to forward the spec for the current instance to `instanceHeader`, check if it's a parsing one or not and add `IOException` only in case if it's a parsing one? This would work the same for all other languages, only the signature of `instanceHeader` would need to be changed.

Some examples of my KSY:

      is_header_none:
        value:  (control_info == 0x50) or
                (control_info == 0x51) or
                (control_info == 0x52) or
                (control_info == 0x69) or
                (control_info == 0x70) or
                (control_info == 0x71) or
                (control_info == 0x78) or
                (control_info == 0x79)

      header_type:
        value: "is_header_none    ? header_type::none   :
                (
                  is_header_short ? header_type::short  : header_type::long
                )"

...and what is generated for Java:

        private Boolean isHeaderNone;
        public Boolean isHeaderNone() throws IOException {
            if (this.isHeaderNone != null)
                return this.isHeaderNone;
            boolean _tmp = (boolean) ( ((controlInfo() == 80) || (controlInfo() == 81) || (controlInfo() == 82) || (controlInfo() == 105) || (controlInfo() == 112) || (controlInfo() == 113) || (controlInfo() == 120) || (controlInfo() == 121)) );
            this.isHeaderNone = _tmp;
            return this.isHeaderNone;
        }

        private HeaderType headerType;
        public HeaderType headerType() throws IOException {
            if (this.headerType != null)
                return this.headerType;
            this.headerType = isHeaderNone() ? HeaderType.NONE : isHeaderShort() ? HeaderType.SHORT : HeaderType.LONG;
            return this.headerType;
        }

GreyCat:
[...]

PR #58, opened 2017-01-05
Less io exceptions

tschoening:
This pull request is related to the issues underneath: The runtime declares to throw a lot of `IOException` while it actually isn't, at least in current versions. The problem with those checked exceptions is that they propagate into users code and need to be handled in many different places, which is unnecessary work using try/catch or unnecessary changes for already available method signatures. Without all those exceptions the JavaCompiler can be cleaned up as well a lot and will only keep one of those exceptions where it actually makes sense.

In the integration task I'm currently facing, where I need to replace an old parser with a KS generated one, this makes a huge difference. And we don't even loose anything, because as said, most of those `IOExceptions` were only declared, not needed at all.

One exception is `processZlib` which replaced another checked exception with `IOException`, which make sense of course to not use too many different exception types. But in other places like `ensureFixedContents", @GreyCat decided to use a custom class extending `RuntimeException` as well, so I think it's reasonable to use such in this case, too.

I've tried various other implementations in the JavaCompiler to reduce the amount of `IOException` to the absolute minimum, but nothing is as easy as simply cleaning up the runtime and afterwards cleaning up the compiler.

https://github.com/kaitai-io/kaitai_struct_compiler/issues/57
https://github.com/kaitai-io/kaitai_struct_java_runtime/issues/3

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
@@ -72,7 +72,7 @@ class JavaCompiler(config: RuntimeConfig, out: LanguageOutputWriter)
 
   override def classConstructorHeader(name: String, parentClassName: String, rootClassName: String): Unit = {
     out.puts
-    out.puts(s"public ${type2class(name)}($kstreamName _io) throws IOException {")
+    out.puts(s"public ${type2class(name)}($kstreamName _io) {")
     out.inc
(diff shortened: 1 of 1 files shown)


GreyCat:
I totally agree that in its current implementation, these IOExceptions can be evaded. And even if we're going to ever re-support file-based I/O, probably it's wise to wrap IOExceptions there into some our custom IO exception, based on RuntimeException.

PR #60, opened 2017-01-09
Process bcd to

tschoening:
PoC for support of BCD coding using "process" for [#78](https://github.com/kaitai-io/kaitai_struct/issues/78).


tschoening:
Supporting BCD using process seems like a dead-end, so closing and deleting my branches. User types should be preferred or maybe external `processes` like discussed in the following:

https://github.com/kaitai-io/kaitai_struct/issues/137#issuecomment-288363610

PR #63, opened 2017-01-27
CalcIntType -> long, because int might lead to loss of data…

### LINKED ISSUE #510: CalcIntType "int" might lead to loss of data (in Java)
Consider the following user type to represent a 6 byte unsigned integer:

    u6le:
      seq:
        - id:   lsb
          type: u1
        - id:   middle
          type: u4
        - id:   msb
          type: u1

And the generated Java:

        private Integer value;
        public Integer value() {
            if (this.value != null)
                return this.value;
            int _tmp = (int) ((((msb() << 40) | (middle() << 8)) | lsb()));
            this.value = _tmp;
            return this.value;
        }
        private int lsb;
        private long middle;
        private int msb;

The cast is wrong in this case, isn't it? As far as I can see, the compiler maps the calculation to the type `CalcIntType` and that is mapped to `int` by convention and that not only in Java, but in other languages like C++ as well. In the most of my cases this is a reasonable choice, but the example above is one exception to the rule. :-)

So shouldn't `CalcIntType` be changed to `long` just to be sure, at least in Java?

Best would be of course that the compiler could check all parts of the calculation and derive the type from the largest, but currently it doesn't seem able to do so.

On the other hand, `long` might waste memory in some cases. But because of AutoBoxing used pretty often in KS, this ins't necessarily too bad. Callers could simply work with the already existing objects instead of primitive types. Forwarding the already available objects from KS shouldn't have any bad influence on memory footprint at all.

ams-tschoening:
…in some calculations.

Resolves https://github.com/kaitai-io/kaitai_struct/issues/510


GreyCat:
Actually, I wanted to return to this topic. I'll try to check if we can make that transition with the tests and hopefully we'll do it. Current "default is 32-bit int" indeed makes things much more complicated than it could be.

PR #64, opened 2017-01-27
doIntLiteral couldn't handle >32 Bit signed int

tschoening:
I had to deal with int literals of less and more than 32 Bit signed int and in the latter case the generated code for Java didn't compile, because the compiler provided the error message that the number is too large to be an int. I needs the suffix "L" in such cases and that is now added for Java only.


GreyCat:
Thanks!

PR #67, opened 2017-02-07
#66: Enhanced ArrayType with the size property…

tschoening:
…so that one can check how many records already have been parsed or how many at all etc. Depends on when this property is accessed by whom.

https://github.com/kaitai-io/kaitai_struct_compiler/issues/66


GreyCat:
Looks good! Thanks!

Issue #65, opened 2017-02-05
Reading strings doesn't handle different character order?


tschoening:
The format I'm currently working on has some byte range which could simply encode textual characters in ISO-8859-1. I have the following sentence in my spec:

> Like all multi byte fields the **last** character is transmitted **first**.

From looking at the Java runtime and the associated tests, this is currently not handled by the string reading functions. Or am I missing something? From my understanding you simply read bytes in order and create a string of them, all in read order of the bytes, which would violate my spec.

So do we need some "order" attribute for the individual characters to define which is the first one? Depending on that the runtime could simply reverse the read string in the easiest case.

It gets problematic for multi-byte encoding,  though.

On the other hand, isn't a string itself in its whole simply multi-byte data and needs to honour the endiannes, like an u8? So shouldn't data markes as little endian with a string always be read as my spec says, last char first? That would violate you current test, though. In theory, each individual char containing of multiple bytes would need to be read like the endianness tells as well?

GreyCat:
Um, actually it doesn't seem to make any sense at all.

ISO8859-1 is a single-byte encoding. There's nothing to swap places and I've never ever seen anyone writing stream backwards (i.e. writing "dcba" to get "abcd"), motivating it with endianness. Technically, if anyone does that, that's not a huge problem to implement some method like `reverse` to reverse either the string or byte array before conversion to a string (hmm, by the way, we're definitely missing byte array -> string conversion method as well).

Things get slightly messier with multi-byte encodings, especially ones that do fixed amounts of bytes (>1) per character. That's UTF16, UTF32, etc. However, the humanity have already solved that problem by using encoding names like "UTF16LE" and "UTF16BE" to differentiate between two possible variants, i.e. "abcd" encoded as:

* `61 00 62 00 63 00 64 00` (UTF16LE)
* `00 61 00 62 00 63 00 64` (UTF16BE)

Note that, again, it does not reverse the string as a whole, it just reverses each multi-byte character in string.

So, if I'd think of improving the situation somehow, the best I can think of is treating encodings like "UTF16" specially by appending "BE" or "LE" suffix automatically based on current default endianness — however, that's not a very good idea either, as many encoding engines treat "UTF16" without BE and LE specially, i.e. expecting [BOM](http://en.wikipedia.org/wiki/Byte_order_mark) — and it's probably a bad idea to break their functionality there.

tschoening:
[...]

PR #69, opened 2017-02-07
Added the operation "reverse" on strings.

tschoening:
Only C++ is not implemented currently, because I don't know how to deal with locally needed temp instances of data. We need something like the following, but I didn't find how to temporarily generate the `str` part.

> std::reverse(str.begin(), str.end());

https://github.com/kaitai-io/kaitai_struct_compiler/issues/65


koczkatamas:
One of the options is to put a helper function into the C++ runtime which reverses the whole string without using the expression twice.

GreyCat:
Looks nice! Thanks!

Issue #78, opened 2017-03-31
condRepeatExprHeader doesn't seem to get a needRaw correctly in case of switch.


tschoening:
This is somewhat related to #77, the compiler seems to fail to generate creation code for the needed `_raw_registers` list. There's only one `new ArrayList` in the following code, which leads to a NPE in Java when raw data should be stored.

        private void _read() {
            [...]
            registers = new ArrayList<Object>(Long.valueOf(registersCnt()).intValue());
            for (int i = 0; i < registersCnt(); i++) {
                switch (spacingControlByte().dataCoding()) {
                case BCD_08_DIGITS: {
                    this._raw_registers.add(this._io.readBytes(spacingControlByte().getFixedDataSize()));
                    [...]
                }

In theory the compiler should handle the raw case already:

	  override def condRepeatExprHeader(id: Identifier, io: String, dataType: DataType, needRaw: Boolean, repeatExpr: expr): Unit = {
	    if (needRaw)
	      out.puts(s"${privateMemberName(RawIdentifier(id))} = new ArrayList<byte[]>
            [...]
	  }

This seems to be a general problem in the compiler, because I get the same error using KSV in Ruby:

> [...]/fmt_oms_rec.rb:2323:in `block in _read': undefined method `[]=' for nil:NilClass (NoMethodError)
>         from C:/Users/tschoening/AppData/Local/Temp/d20170331-1608-u0r8oz/fmt_oms_rec.rb:2238:in `times'
>         from C:/Users/tschoening/AppData/Local/Temp/d20170331-1608-u0r8oz/fmt_oms_rec.rb:2238:in `_read'
>         from C:/Users/tschoening/AppData/Local/Temp/d20170331-1608-u0r8oz/fmt_oms_rec.rb:1070:in `_read'

The compiler generated only one `Array.new` as well:

    def _read
      [....]
      @registers = Array.new(registers_cnt)
      (registers_cnt).times { |i|
        (@_debug['registers'][:arr] ||= [])[i] = {:start => @_io.pos}
        case spacing_control_byte.data_coding
        when :fmt_oms_apl_vdb_header_dib_data_coding_bcd_08_digits
          (@_debug['registers'][:arr] ||= [])[i] = {:start => @_io.pos}
          @_raw_registers[i] = @_io.read_bytes(spacing_control_byte.get_fixed_data_size)
          [....]

This looks like `needRaw` is not properly calculated for my `switch` with `repeat` for some reason:

[...]

PR #79, opened 2017-03-31
needRaw didn't properly work for SwitchType...

tschoening:
...which might consist of different types for all cases. It should be sufficient to test if one of those cases fullfills needRaw to get the overall result for needRaw for the whole switch.

https://github.com/kaitai-io/kaitai_struct_compiler/issues/78

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/components/EveryReadIsExpression.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/components/EveryReadIsExpression.scala
@@ -200,7 +200,7 @@ trait EveryReadIsExpression extends LanguageCompiler with ObjectOrientedLanguage
 
   def needRaw(switchType: SwitchType): Boolean = {
     val byCase = switchType.cases.mapValues(v => needRaw(v))
-    val byBool = byCase.groupBy(pair => pair._2)
+    val byBool = byCase.groupBy({ case (k, v) => v })
     val retVal = byBool.contains(true)
 
     retVal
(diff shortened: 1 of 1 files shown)


tschoening:
Closing this because it has been fixed in the meanwhile in #205 with the following commit:

https://github.com/kaitai-io/kaitai_struct_compiler/commit/a7f1167717c15ba3ce1480d67c4cab2eacfad196#diff-3b9e2ea7df1c77e6702738c198f9cd35R84

This saves me from resolving the merge conflict in #79 as well. I will document my little bit different approach for fixing this here, though, so I can delete my local branch entirely while still keeping the contents, just in case...

	def needRaw(dataType: DataType): Boolean = {
		dataType match {
			case _: UserTypeFromBytes => true
			case st: SwitchType => st.hasSize
			case _ => false
		}
	}

vs.

	def needRaw(dataType: DataType): Boolean = {
		dataType match {
			case t: UserTypeFromBytes => true
			case t: SwitchType => needRaw(t)
			case _ => false
		}
	}

	def needRaw(switchType: SwitchType): Boolean = {
		val byCase = switchType.cases.mapValues(v => needRaw(v))
		val byBool = byCase.groupBy({ case (k, v) => v })
		val retVal = byBool.contains(true)

		retVal
	}

GreyCat:
Thanks for original work and revisiting it for update! I'm really sorry for keeping it stalled, hopefully I'll try to nail other issues for you in timely manner ;)

Issue #80, opened 2017-04-03
Wrong output for ksc-json-output in case of relative paths under Windows.


tschoening:
I upgraded `ksv` to get support for [opaque types](https://github.com/kaitai-io/kaitai_struct_visualizer/issues/15) and with it came a changed interaction with `ksc`, the new `ksc-json-output` is used now. This leads to compile time errors in `ksv` for me, because your generated JSON doesn't seem to handle paths and their different representation in Java under different OS correctly. 

> C:/[...]/visualizer_main.rb:68:in `block (2 levels) in compile_formats': undefined method `[]' for nil:NilClass (NoMethodError)
        from C:/Program Files (x86)/kaitai_struct/visualizer/lib/kaitai/struct/visualizer/visualizer_main.rb:64:in `each'
        from C:/Program Files (x86)/kaitai_struct/visualizer/lib/kaitai/struct/visualizer/visualizer_main.rb:64:in `each_with_index'
        from C:/Program Files (x86)/kaitai_struct/visualizer/lib/kaitai/struct/visualizer/visualizer_main.rb:64:in `block in compile_formats'

``` Ruby
      fns.each_with_index { |fn, idx|
        puts "... processing #{fn} #{idx}"

        log_fn = log[fn]
        if log_fn['errors']
```

``` JSON
{"record\fmt_clt_recs.ksy": {"firstSpecName": "fmt_clt_recs","output": {"ruby": {"fmt_clt_recs": {"topLevelName": "FmtCltRecs","files": [{"fileName": "fmt_clt_recs.rb"}]},"fmt_oms_rec": {"topLevelName": "FmtOmsRec","files": [{"fileName": "fmt_oms_rec.rb"}]}}}},"record\fmt_oms_rec.ksy": {"firstSpecName": "fmt_oms_rec","output": {"ruby": {"fmt_oms_rec": {"topLevelName": "FmtOmsRec","files": [{"fileName": "fmt_oms_rec.rb"}]}}}}}
```

There are two problems here:

1. / to \

I use a command line similar to the following in Eclipse to start `ksv`:

> [...]ksv "${workspace_loc}/[...]/data.bin" "record/fmt_clt_recs.ksy" "record/fmt_oms_rec.ksy"

[...]

PR #81, opened 2017-04-03
Normalize path handling, so that / only is used.

tschoening:
Makes interaction with ksv easier and fixes escape problems with `\`.

https://github.com/kaitai-io/kaitai_struct_compiler/issues/80


tschoening:
@GreyCat decided for a more cleaner solution, so this PR is not needed anymore.

PR #93, opened 2017-06-28
JavaScriptCompiler: fix debug output

tecknicaltom:
In the calculated endianess branch, the compiler was changed to always
emit the _read function instead of only when debug was enabled. The
readHeader method was introduced to emit the header of this function.
The JavaScriptCompiler still had code in the classConstructorHeader
method to emit the _read header if debug.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/JavaScriptCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/JavaScriptCompiler.scala
@@ -100,14 +100,8 @@ class JavaScriptCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
 
     if (debug) {
       out.puts("this._debug = {};")
-      out.dec
-      out.puts("}")
-      out.puts
-      out.puts(s"${type2class(name.last)}.prototype._read = function() {")
-      out.inc
-    } else {
-      out.puts
     }
+    out.puts
   }
 
   override def classConstructorFooter: Unit = {


GreyCat:
Good catch, thanks!

GreyCat:
However, this patch actually breaks debug mode. It generates:

```javascript
  function Debug0(_io, _parent, _root) {
    this._io = _io;
    this._parent = _parent;
    this._root = _root || this;
    this._debug = {};

    this._read();
  }
  Debug0.prototype._read = function() {
```

i.e. with a `_read()` call, which there shouldn't be one — the point of debug mode is being able to call it manually.

GreyCat:
And, yeah, the same problem actually happens in Ruby, which is incidentally also saved by allowing method redefinitions:

```ruby
  def initialize(_io, _parent = nil, _root = self)
    super(_io, _parent, _root)
    @_debug = {}
  end

  def _read
    _read

    self
  end
  def _read
    (@_debug['one'] ||= {})[:start] = @_io.pos
    # ...
```

GreyCat:
Ok, looks like I've messed a lot of stuff up. Back to [drawing board](https://github.com/kaitai-io/kaitai_struct/issues/17).

tecknicaltom:
ah good catch. I wasn't sure if the call to _read was supposed to be in the ctor, but when comparing debug vs non-debug, it seemed to me like it was intentional. I didn't look at ruby (I'm in the process of trying to implement debug support for perl and python)

GreyCat:
If you're interested in Perl — may be you could take a look at fixing broken tests in CI for Perl (there's something related to array element getting)? It looks like a guy who supported Perl compiler for us disappeared recently :(

PR #94, opened 2017-07-02
fix several issues with perl compiler

tecknicaltom:
* add a case in doLocalName for INDEX to $i. Otherwise, it returned
$self->{$i}. This fixes the error in IndexSizes test
* Perl doesn't support an indexing directly on the return value of a
function (e.g. foo()[3]), and since we always deal with array refs, in
doSubscript, always explicitly dereference the container as a list with
circumfix operator (e.g. @{foo()}[3]). This fixes the SwitchCast test


GreyCat:
Thanks! Off to the CI it goes :)

PR #95, opened 2017-07-04
Perl: add import of List::Util when needed

tecknicaltom:
Should fix the CI error "Undefined subroutine &List::Util::min called at
compiled/perl/ExprArray.pm line 79."


GreyCat:
Thanks again!

PR #96, opened 2017-07-05
Add ObjectOrientedLanguage.localTemporaryName

tecknicaltom:
In the process of supporting debug mode for Perl, but wanted sign-off on
this before proceeding. Previously,
EveryReadIsExpression.attrUserTypeParse generated the name of the
temporary variable as s"_t_${idToStr(id)}" but this won't work for Perl
because it lacked the sigil. This change introduces a new method called
localTemporaryName to ObjectOrientedLanguage to allow for
language-specific generation of the name of a temporary variable meant
to store a value for a given Identifier.

If you're not happy with this way of addressing the issue, please provide some guidance on how I should proceed. Thanks!

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala
@@ -521,6 +521,8 @@ class CSharpCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
       case _ => s"_${idToStr(id)}"
     }
   }
+
+  override def localTemporaryName(id: Identifier): String = s"_t_${idToStr(id)}"
 }
 
 object CSharpCompiler extends LanguageCompilerStatic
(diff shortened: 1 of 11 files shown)


GreyCat:
This is a very specific area, so it is actually only used in `handleAssignmentTempVar`. Why don't we just add `$` to pre-generated `_t_whatever` in Perl's implementation of `handleAssignmentTempVar`?

In many cases, customization of just the name of the variable (whether it would include or not include any `$`, `@`, `self.`, `this.` or any other special suffixes/prefixes) is not enough. For example, in C you would need to put extra effort into registering that local variable in some separate stream, I guess.

tecknicaltom:
the temporary variable is passed to three functions:

1. ``handleAssignmentTempVar`` - only used for this one case, would be fine to prepend the sigil in the perl version
1. ``userTypeDebugRead`` - also used by the NoRepeat case. In that case, ``userTypeDebugRead`` is given the output of ``privateMemberName`` with the sigil included. having this function prepend or not prepend the sigil seems problematic
1. ``handleAssignment`` - a pretty generic function that wouldn't add a sigil to the potentially arbitrary ``expr`` passed to it

I guess the next cleanest implementation I can think of is to have ``handleAssignmentTempVar`` generate and return the temporary variable name, but it seemed like that was having one function serve two purposes which I thought you'd prefer to avoid.

GreyCat:
Yeah, that's decent reasoning, thanks!

PR #100, opened 2017-08-13
GoCompiler: fix handleAssignmentRepeatExpr

nolta:
Fixes the "fix" in 05579b8c4ac0184b8a3ce192bf91234a81139316.


GreyCat:
Thanks again! I wonder if you'd want direct commit rights into the compiler and/or tests repos? All your changes seem to be pretty good :)

PR #106, opened 2017-09-02
#241 fixed: true.to_i becomes now (true)?1:0 to both avoid errors due…

tempelmann:
… to missing parentheses and avoid warnings due to using bool values in an int expression.


GreyCat:
Thanks! Makes sense, but could you please make that `) ? 1 : 0)"` — i.e. with spaces? Without spaces, it fails just about any C++ style guide I can think of.

tempelmann:
Feel free to just edit the spaces right in the pull requests.

GreyCat:
Done that, off to CI it goes.

PR #108, opened 2017-09-02
#250: Uses "not" and "and" instead of "!" and "&&" operators

tempelmann:


GreyCat:
Given that the original idea haven't got much favor in the discussion, I'm closing this one without merging. If someone really wants it, it still can added as *alternative*, probably enabled by some sort of CLI option.

PR #111, opened 2017-09-02
#252 - groups the public and private class members.

tempelmann:
Makes the .h file neat and tidy :)


tempelmann:
Note that I "abused" the ImportList for collecting the lines. I didn't want to create yet another class with the same functionality.

tempelmann:
About the bot warnings: I should probably declare them private.

Also: It's still buggy: Comments (from "doc:") do not generate correctly.

tempelmann:
The comments bug comes from the fact that I only add unique lines - a mistake, when it comes to comment lines. Fixing that now...

tempelmann:
Alright. Fixed.

tempelmann:
I can't avoid the mutable warning, though. I have to change the `publicDecls` between classHeader and classFooter, to handle nesting of classes. No idea how to do that otherwise.

PR #122, opened 2018-01-18
python translator: b"..." literals instead of struct.pack

arekbulski:
I dont speak Scala so this is guess-work code for literals.


GreyCat:
LGTM :)

PR #123, opened 2018-01-30
Update README.md, add missing target languages

andreaskern:
`go` and `graphviz` were missing in the list, the order is the same as in the `--help` message


GreyCat:
Uh. Actually, we also need to fix order in `--help`, as right now it's unstable %) Thanks!

PR #125, opened 2018-02-08
Fix URL to Web IDE

kylestev:
The target URL used a `/` in place of a `.` causing the TLD that is looked up to be `kaitai` which is not a TLD.


kylestev:
Case in point:

GreyCat:
Thanks!

PR #131, opened 2018-02-27
Access foreign types

tschoening:
I started to have  a look at [#275](https://github.com/kaitai-io/kaitai_struct/issues/275) and as things seems to get more difficult than expected, I would like to provide what I have so far.


GreyCat:
Thanks!

PR #135, opened 2018-03-30
Remove warning about unused i in Java.

tschoening:
This fixes https://github.com/kaitai-io/kaitai_struct/issues/365 in a somewhat easy manner and can be improved in future if needed at all.


tschoening:
Do you really think one will have such a detailed look at the generated code in future? I don't unless I have problems or Eclipse marks my project with a warning sign. And currently there's no way to get rid of that sign unless one manually removes the usage of `i` after each and every new generation of code. I have 3 places in two files already and that started to annoy me.

Add a `FIXME` like in other cases already, linking to the issue. You can even output in in the generated Java code, so everyone with such a detailed look at the generated code can directly start working on a cleaner solution. ;-)

PR #137, opened 2018-04-04
Complete Expressions tests for go

cugu:


jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
```diff
@@ -119,6 +119,7 @@ class TranslatorSpec extends FunSuite {
   full("some_bool.to_i", CalcBooleanType, CalcIntType, Map[LanguageCompilerStatic, String](
     CppCompiler -> "some_bool()",
     CSharpCompiler -> "(SomeBool ? 1 : 0)",
+    GoCompiler -> "func()(int) { if ( this.SomeBool ) { return 1 }; return 0 }()",
```
GreyCat: There is actually no need to wrap that in such anonymous inline function. We can render Go expressions with multiple-statement output with temporary variables, i.e. as ternary operator gets rendered.

jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
```diff
@@ -156,6 +159,7 @@ class TranslatorSpec extends FunSuite {
   full("foo.bar", FooBarProvider, CalcStrType, Map[LanguageCompilerStatic, String](
     CppCompiler -> "foo()->bar()",
     CSharpCompiler -> "Foo.Bar",
+    GoCompiler -> "this.foo.bar",
```
GreyCat: I believe it's `this.Foo.Bar`?

GreyCat:
Thanks! I guess we'll merge it now and fix stuff later ;)

PR #138, opened 2018-04-06
Improve GoTranslator

cugu:


jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
```diff
@@ -425,7 +425,7 @@ class TranslatorSpec extends FunSuite {
   full("\"str\".reverse", CalcIntType, CalcStrType, Map[LanguageCompilerStatic, String](
       CppCompiler -> "kaitai::kstream::reverse(std::string(\"str\"))",
       CSharpCompiler -> "new string(Array.Reverse(\"str\".ToCharArray()))",
-      GoCompiler -> "func(s string)(string)(r := []rune(s); for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 { r[i], r[j] = r[j], r[i] }; return string(r))(\"str\")",
+      GoCompiler -> "func(s string)(string){r := []rune(s); for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 { r[i], r[j] = r[j], r[i] }; return string(r)}(\"str\")",
```
GreyCat: Shall we just put it into runtime, please?

cugu: Sure, we can do that.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -237,6 +237,9 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     "big5" -> ("traditionalchinese.Big5", "golang.org/x/text/encoding/traditionalchinese")
   )
 
+  override def bytesToStr(value: Ast.expr, expr: Ast.expr): TranslatorResult =
+    bytesToStr(translate(value), expr)
```
GreyCat: Um, isn't it an infinite recursion?

cugu: I don't know. I don't know scala :-D Isn't it just method overloading like here: https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala#L154

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -304,28 +307,77 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
...
+    ResultString(s"strconv.FormatInt(int64(${translate(value)}), ${translate(base)})")
+  }
 
   override def floatToInt(value: Ast.expr): TranslatorResult =
     ResultString(s"int(${translate(value)})")
 
-  override def kaitaiStreamSize(value: Ast.expr): TranslatorResult = ???
+  override def kaitaiStreamSize(value: Ast.expr): TranslatorResult = {
+    ResultString(s"${translate(value)}.Size()")
+  }
```
GreyCat: As far as I understand, this one is able to "throw an exception", thus it returns not just a value, but value + error. So, for expression translation purposes, its needs to assign stuff a temporary variable + err variable, check the err variable, and return temp var as result. [outVarCheckRes](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala#L351-L356) already does that — take a look, for example, at how [bytesToStr](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala#L240-L260) uses it.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -304,28 +307,77 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
...
 
-  override def kaitaiStreamEof(value: Ast.expr): TranslatorResult = ???
+  override def kaitaiStreamEof(value: Ast.expr): TranslatorResult = {
+    ResultString(s"${translate(value)}.EOF()")
+  }
 
-  override def kaitaiStreamPos(value: Ast.expr): TranslatorResult = ???
+  override def kaitaiStreamPos(value: Ast.expr): TranslatorResult = {
+    ResultString(s"${translate(value)}.Pos()")
+  }
```
GreyCat: I believe that `kaitaiStreamEof` and `kaitaiStreamPos` also can "throw exception" and thus require err check.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -304,28 +307,77 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
...
+    out.inc
+    out.puts(s"if ${localVarName(max)} < ${localVarName(value)} {")
+    out.inc
+    out.puts(s"${localVarName(max)} = ${localVarName(value)}")
+    out.dec
+    out.puts("}")
+    out.dec
+    out.puts("}")
+    ResultLocalVar(max)
+  }
```
GreyCat: I'm kind of hesitating about these, as there's 6 lines to be produces, but as far as I understand, if we want to put into runtime, that would require runtime to have tons of methods for any possible array type, right?

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
GreyCat: Right, sorry, my bad. Definitely need to have more sleep ;)

GreyCat:
LGTM! Let's take it for a spin and see what will happen ;)

PR #140, opened 2018-08-03
Changed the dependency in the file

KOLANICH:


GreyCat:
Please provide your rationale. This particular setting was chosen after a long discussion in sbt.

KOLANICH:
The rationale not to install an obsolete version of java to modern versions of distros. IMHO this case is itself too restrictive: there are impls of java having other package name. Could you give me a link to the discussion you have meant?

GreyCat:
Contrary to what it might suggest, `java8-runtime-headless` does not install "obsolete version of java", but rather requires any JRE 8+ to be present on the system. I.e. if you have either Oracle's or OpenJDK build of Java 10, it should work and not force you to install actual JRE 8.

Here's the discussion: https://github.com/sbt/sbt-native-packager/issues/1106

KOLANICH:
That's nice. Thank you for the info.

PR #141, opened 2018-08-03
Added some manipulation of env vars

KOLANICH:
Not very familiar  (completely unfamiliar) with scala and sbt, the code may be wrong.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -118,8 +118,11 @@ lazy val compiler = crossProject.in(file(".")).
 
     // Remove all "maintainer scripts", such as prerm/postrm/preinst/postinst: default
     // implementations create per-package virtual user that we won't use anyway
-    maintainerScripts in Debian := Map(),
-
+    maintainerScripts in Debian := Map(
+      Postinst -> "echo \"KAITAI_STRUCT_ROOT=/usr/share/kaitai-struct-compiler\" >> /etc/environment",
+      Postrm -> "sed -ir -e \"/^KAITAI_STRUCT_ROOT=.\\+$/d\" /etc/environment"
+    ),
+    
     // Work around new Debian defaults and sbt-native-packager defaults, which
     // build .deb packages that appear to be incompatible with older Debian/Ubuntu's
     // dpkg and are not accepted by BinTray.


KOLANICH:
>I don't think this is how Debian people update their environment.

What do you think is a better already-existing way?

>Why do you want to do that?

Because kaitaiStructCompile relies on this env variable.

GreyCat:
> What do you think is a better already-existing way?

I believe supplying a file in `/etc/profile.d` is the current Debian/Ubuntu way.

Although I still want to emphasize that setting an system-wide environment variable is a bad idea and I don't see any particular scenario where it will help.

> Because kaitaiStructCompile relies on this env variable.

Ok, let's recap this. ksc has never included anything like `KAITAI_STRUCT_ROOT`. You've invented a variable and want exactly one (.deb) of many different builds of ksc (.zip, .msi, staging, etc) to include it. No rationale was provided on:

* What it does
* How it acts
* What it's used for
* Why it has to be exactly environment variable

Do you really expect me to accept this?

I believe this is a [XY problem](http://xyproblem.info/). You don't really want a variable per se, but there's something you want to be solved?

KOLANICH:
>I believe this is a XY problem. You don't really want a variable per se, but there's something you want to be solved?

`kaitaiStructCompile` in order to work needs paths to KSC, since KSC may be not installed into the system and/or may be not available in `PATH`.

So it needs:
* a path to KSC is needed if the module uses command line intercace (the most robust and the same way ideologically incorrect and fragile solution (there were cases of vulnrs when people used cli where they should use API) )
* a path to jars if it uses `jpype` (a module allowing to host a JVM inside of cpython process and call into from python into java (and any other jvm languages too with a bit of perversions ) ) or similar solutions allowing calling java libs from python (jython, graalvm) to use in a more sensible, secure and hackable maner.

So, we need to pass paths. I think it makes sense to put it ino an env variables, especially taking in account that this is the common approach, lot of software, like Visual Studio, Java, gnu Octave, do the same.  Fortunately for both linux and windows editions there exists a common prefix, so we can use the single variable.

KOLANICH:
>You've invented a variable and want exactly one (.deb) of many different builds of ksc (.zip, .msi, staging, etc) to include it.

MSI should also include, I have only chosen to implement setting of it for deb first.

GreyCat:
> `kaitaiStructCompile` in order to work needs paths to KSC, since KSC may be not installed into the system and/or may be not available in `PATH`.

Actually, in 100% of the system-wide installations I know (i.e. Debian package, OS X brew, Windows MSI), compiler is guaranteed to be in `PATH`. In all other cases, it makes perfect sense to ask user to add `kaitai-struct-compiler` into `PATH` manually.

> a path to jars if it uses `jpype` (a module allowing to host a JVM inside of cpython process and call into from python into java (and any other jvm languages too with a bit of perversions ) ) or similar solutions allowing calling java libs from python (jython, graalvm) to use in a more sensible, secure and hackable maner.

Ok, in this case it basically needs to duplicate the logic of `kaitai-struct-compiler` launcher script. And it is actually relatively simple too: once you've located `kaitai-struct-compiler` and resolved full path to it, you can just go one level up in directories and `lib/*.jar` would be what you want.

In Debian, for example, it works like that:

* /usr/bin/ksc
* symlink resolved to /usr/bin/kaitai-struct-compiler
* symlink resolved to /usr/share/kaitai-struct-compiler/bin/kaitai-struct-compiler
* derived as /usr/share/kaitai-struct-compiler/lib/*.jar

Alternatively, if you intend to use jars as a real JVM-based application would use, you would probably declare a dependency on it using maven/ivy-like syntax, pull proper jar + dependencies somehow using standard Java depency resolution methods (again, like maven or ivy downloads) and load jars from its jar cache. My best bet would be that `jpype` would support something like that. If yes, you should just depend on [kaitai-struct-compiler](https://oss.sonatype.org/content/repositories/snapshots/io/kaitai/kaitai-struct-compiler_2.11/).

GreyCat:
> I dislike this solution since having this info in an env variable is imho better.

It's not. Generally, introducing a mandatory environment variable is a huge pain in the ass. For example, in typical Unix/Linux, you can't just install the package and use it right away: you'll have to restart your shell and, in worst case, relogin, to get these environments applied. On Windows, in best case it's possible to push environment update everywhere, but it worst case scenario (with weird group policies in place or if your system has lots of "All users" vs "Local machine" vs "Current user" intricacies), it becomes much more cryptic.

For an average user, setting environment variables (and fixing misbehaving setups) is relatively hard and unintuitive.

Stuff like `JAVA_HOME`, `OCTAVE_HOME`, etc is mostly invented long ago and out of desperation due to lack of other options.

KOLANICH:
OK, thank you.

PR #142, opened 2018-08-09
Partial implementation of Rust output

CWood1:
This is an (as yet incomplete) implementation of Rust output, as discussed in issue https://github.com/kaitai-io/kaitai_struct/issues/22.


shared/src/main/scala/io/kaitai/struct/languages/components/LanguageCompiler.scala
```diff
@@ -116,7 +116,8 @@ abstract class LanguageCompiler(
   def seek(io: String, pos: Ast.expr): Unit
   def popPos(io: String): Unit
   def alignToByte(io: String): Unit
-
+
```
GreyCat: This one adds trailing whitespaces.

GreyCat:
Merging this in as well. Thanks you very much for all this hard work! It's very impressive!

PR #147, opened 2018-10-23
WIP: remove src folder from generated structure

Vineg:
You don't always want sources to be generated in /src folder, especially when you put --outdir parameter. It can be helpful if you want files to be directly generated to the project with project structure like `...src/java/io/kaitai/...`
If you approve, we should probably remove src from other languages as well.


Vineg:
Note I am not the only one who noticed https://github.com/jeremywall-davis/kaitai_struct_compiler/commit/11f023098f0cfbea2048c8517661f37289156537

GreyCat:
Sorry for late reply.
Yeah, this is somewhat stupid problem, which boils down to solving two things:

1. This is breaking change, i.e. it will break stuff for people who currently rely on generation of exactly that path. At the bare minimum, we'll need to document it in release notes.
2. Currently, this breaks tests, so we'll need to address that too.

I can't merge this in until these two problems will be addressed. Let's see what I can do about that.

PR #156, opened 2019-01-27
Added a command to build my docker image

KOLANICH:


GreyCat:
Ok, two major questions here:

1. What is it at all, why do we need this? After some digging, I've found out that it's project https://gitlab.com/kaitaiStructCompile.py/kaitai_struct_python_docker that you want to trigger rebuild, I guess, but it's still pretty vague what is .
2. What relation does it have with bintray .deb publication? If there's none, and it's just another action to do during the build — it should be a clearly separate script, getting separate invocation from .travis.yml and/or circleCI.

KOLANICH:
>What is it at all, why do we need this? After some digging, I've found out that it's project https://gitlab.com/kaitaiStructCompile.py/kaitai_struct_python_docker that you want to trigger rebuild, I guess, but it's still pretty vague what is .

It is a Docker image with everything needed to build python projects using KS. Needed because prebuilt docker images are faster than building environment from scratch every time.

>What relation does it have with bintray .deb publication? 

It uses it. The plan is: when a new unstable version is published, my scripts are triggered and rebuild the image, fetching the deb and installing it. In fact when Alpine package is ready, it will use it instead of deb, because Alpine images are smaller.

GreyCat:
Ok, understood. I won't merge this PR exactly, but I would rather create a separate step on travis.yml level, that will run if .deb publication succeeded. Please also note that actually [sbt-native-packager](https://github.com/sbt/sbt-native-packager) that we use also can create Docker images & GraalVM native images directly, so that might be something that you'd like to look into.

GreyCat:
Here you go: https://github.com/kaitai-io/kaitai_struct/commit/b05d31e8a8745c146df44576ef5a944c0e85dcee — hopefully should pass in 10-15 minutes.

KOLANICH:
Thank you, [that worked](https://gitlab.com/kaitaiStructCompile.py/kaitai_struct_python_docker/-/jobs/151870769).

KOLANICH:
>Please also note that actually sbt-native-packager that we use also can create Docker images & GraalVM native images directly, so that might be something that you'd like to look into.

Yes, this stuff definetily should be helpful in order to create official images.

PR #158, opened 2019-04-04
Allow "import-path" more than once to not necessarily use OS-specific…

ams-tschoening:
… path separator. The used args parser already supports everything needed, so this should be the easiest solution.

https://github.com/scopt/scopt#occurrence
This fixes https://github.com/kaitai-io/kaitai_struct/issues/557.


GreyCat:
Merging in, thanks again for detailed investigation and a fix!

PR #160, opened 2019-04-08
CalcArrayType needs to be handled like ArrayType at some places…

ams-tschoening:
…or things might break e.g. in case of value instances returning lists of user types.

This fixes https://github.com/kaitai-io/kaitai_struct/issues/561.


ams-tschoening:
Any interest to review/merge?

ams-tschoening:
Ping

shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
```diff
@@ -789,7 +789,7 @@ object JavaCompiler extends LanguageCompilerStatic
       case t: UserType => types2class(t.name)
       case EnumType(name, _) => types2class(name)
 
-      case ArrayType(_) => kaitaiType2JavaTypeBoxed(attrType)
+      case ArrayType(_) | CalcArrayType(_) => kaitaiType2JavaTypeBoxed(attrType)
```
GreyCat: This probably warrants a bigger change, i.e. refactoring type system to introduce something like common ArrayType ancestor, so we can derive current ArrayType + CalcArrayType from these two, but let me apply your change as a first step.

ams-tschoening: Make sense, especially because of the following:

https://github.com/kaitai-io/kaitai_struct_compiler/pull/160/commits/425da90994cc7e7144c987d7afeb3219921c6c77#diff-d07f3f07809a96c7c4d2db4f02777fb2R163

But I won't have the time to implement such changes. So thanks for merging first!

PR #161, opened 2019-04-18
[WIP] Output for Rust

bspeice:
Alright, been making a lot of progress, but hit some issues I could use help understanding.

Right now, fields/attributes/datatypes have no idea whether they're related to an instance or not. This causes issues because we want to store things as `Option<ActualType>` which is a different type than `ActualType`. C++/Java get around this by treating the instance storage as `null`, and C++ tracks a separate `f_instance_name` to figure out whether it's been assigned to or not. In Rust, we'd really rather not do that because of safety. Instead, using the implemented `struct.instance_name()` call is the right way to do it, but there's (to my knowledge) no way to distinguish between when we should use `instance_name` or `instance_name()`.

As a quick example from the [websockets](https://github.com/kaitai-io/kaitai_struct_formats/blob/master/network/websocket.ksy) schema:

```rust
impl<'a> KaitaiStruct<'a> for WebsocketDataframe<'a> {
    type Parent = Websocket<'a>;
    type Root = Websocket<'a>;

    fn new(
        _parent: Option<&'a Self::Parent>,
        _root: Option<&'a Self::Root>
    ) -> kaitai_runtime::Result<'a, Self>
    where
        Self: Sized,
    {
        let mut s: Self = Default::default();
        s._parent = _parent;
        s._root = _root;
        Ok(s)
    }

    fn read<S: KaitaiStream>(
        &mut self,
        stream: &mut S,
    ) -> kaitai_runtime::Result<'a, ()>
    where
        Self: Sized
    {
        self.finished = stream.read_bits_int(1)? != 0;
        self.reserved = stream.read_bits_int(3)?.into()?;
        self.opcode = stream.read_bits_int(4)?.into()?;
        self.is_masked = stream.read_bits_int(1)? != 0;
        self.len_payload_primary = stream.read_bits_int(7)?.into()?;
        stream.align_to_byte();
        if self.len_payload_primary == 126 {
            self.len_payload_extended_1 = stream.read_u2be()?.into()?;
        }
        if self.len_payload_primary == 127 {
            self.len_payload_extended_2 = stream.read_u4be()?.into()?;
        }
        if self.is_masked {
            self.mask_key = stream.read_u4be()?.into()?;
        }
        if self._root.dataframes[0].opcode != WebsocketOpcode::Text {
            // Two issues here: We don't know that `len_payload` is an instance
            // and thus don't properly call `len_payload()` to initialize.
            // Also, because self.len_payload may be uninitialized, we can't just add
            // a `read_bytes(Option<u64>)` method.
            self.payload_bytes = stream.read_bytes(self.len_payload)?;
        }
        if self._root.dataframes[0].opcode == WebsocketOpcode::Text {
            self.payload_text = panic!("Unimplemented encoding for bytesToStr: {}", "UTF-8");
        }
    }
}

impl<'a> WebsocketDataframe<'a> {
    fn len_payload(&mut self) -> i32 {
        if let Some(x) = self.len_payload {
            return x;
        }

        // This is illegal right now, but not hard to surround with `Some()` to fix.
        self.len_payload = if self.len_payload_primary <= 125 { self.len_payload_primary } else { if self.len_payload_primary == 126 { self.len_payload_extended_1 } else { self.len_payload_extended_2}};
        return self.len_payload.unwrap();
    }
}
```

Is there a good way to handle this? I don't think resorting to `unsafe` and using the C++-style solution is OK.

EDIT: I'm also working on a separate runtime library.


GreyCat:
Cc @CWood1 to take a look?

bspeice:
One other note: the same issue shows up for enums, I'm currently working around it by using `try_into` to let Rust pick the appropriate type.

bspeice:
Making a ton of progress. There's one last borrow checker issue, some polish needed, and then it should be in a presentable state. I've been using Websockets as the example to compile against, and here's where we're at now:

```rust
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

#[allow(unused_imports)]
use kaitai::{self, KError, KResult, KStream, KStruct, KStructUnit};
#[allow(unused_imports)]
use std::convert::{TryFrom, TryInto};
#[allow(unused_imports)]
use std::vec::Vec;

/// The WebSocket protocol establishes a two-way communication channel via TCP.
/// Messages are made up of one or more dataframes, and are delineated by
/// frames with the `fin` bit set.
#[derive(Default, Debug)]
struct Websocket<'a> {
    pub dataframes: Vec<WebsocketDataframe<'a>>,
    _root: Option<&'a Self>,
    _parent: Option<&'a KStructUnit<'a>>,
}
impl<'a> KStruct<'a> for Websocket<'a> {
    type Parent = KStructUnit<'a>;
    type Root = Self;

    fn new(_parent: Option<&'a KStructUnit<'a>>, _root: Option<&'a Self>) -> KResult<'a, Self> {
        let s = Websocket {
            _parent,
            _root,
            ..Default::default()
        };
        Ok(s)
    }
    fn read<S: KStream>(&'a mut self, stream: &mut S) -> KResult<'a, ()> {
        // condRepeatUntilHeader(NamedIdentifier(dataframes), stream, UserTypeInstream(List(dataframe),None,List()), false, Attribute(Name(identifier(_)),identifier(finished)))
        let mut tmpa = WebsocketDataframe::new(Some(self), self._root)?;
        tmpa.read(stream);
        self.dataframes.push(tmpa);
        // condRepeatUntilFooter(NamedIdentifier(dataframes), stream, UserTypeInstream(List(dataframe),None,List()), false, Attribute(Name(identifier(_)),identifier(finished)))
        Ok(())
    }
}
impl<'a> Websocket<'a> {
}
#[derive(Debug, PartialEq)]
enum WebsocketOpcode {
    Continuation,
    Text,
    Binary,
    Reserved3,
    Reserved4,
    Reserved5,
    Reserved6,
    Reserved7,
    Close,
    Ping,
    Pong,
    ReservedControlB,
    ReservedControlC,
    ReservedControlD,
    ReservedControlE,
    ReservedControlF,
}
impl TryFrom<u64> for WebsocketOpcode {
    type Error = KError<'static>;
    fn try_from(flag: u64) -> KResult<'static, WebsocketOpcode> {
        match flag {
            0 => Ok(WebsocketOpcode::Continuation),
            1 => Ok(WebsocketOpcode::Text),
            2 => Ok(WebsocketOpcode::Binary),
            3 => Ok(WebsocketOpcode::Reserved3),
            4 => Ok(WebsocketOpcode::Reserved4),
            5 => Ok(WebsocketOpcode::Reserved5),
            6 => Ok(WebsocketOpcode::Reserved6),
            7 => Ok(WebsocketOpcode::Reserved7),
            8 => Ok(WebsocketOpcode::Close),
            9 => Ok(WebsocketOpcode::Ping),
            10 => Ok(WebsocketOpcode::Pong),
            11 => Ok(WebsocketOpcode::ReservedControlB),
            12 => Ok(WebsocketOpcode::ReservedControlC),
            13 => Ok(WebsocketOpcode::ReservedControlD),
            14 => Ok(WebsocketOpcode::ReservedControlE),
            15 => Ok(WebsocketOpcode::ReservedControlF),
            _ => Err(KError::UnknownEnum(flag)),
        }
    }
}

#[derive(Default, Debug)]
struct WebsocketDataframe<'a> {
    pub finished: bool,
    pub reserved: u64,
    pub opcode: Option<WebsocketOpcode>,
    pub is_masked: bool,
    pub len_payload_primary: u64,
    pub len_payload_extended_1: u16,
    pub len_payload_extended_2: u32,
    pub mask_key: u32,
    pub payload_bytes: &'a [u8],
    pub payload_text: &'a str,
    _root: Option<&'a Websocket<'a>>,
    _parent: Option<&'a Websocket<'a>>,
    len_payload: Option<i32>,
}
impl<'a> KStruct<'a> for WebsocketDataframe<'a> {
    type Parent = Websocket<'a>;
    type Root = Websocket<'a>;

    fn new(_parent: Option<&'a Websocket<'a>>, _root: Option<&'a Websocket<'a>>) -> KResult<'a, Self> {
        let s = WebsocketDataframe {
            _parent,
            _root,
            ..Default::default()
        };
        Ok(s)
    }
    fn read<S: KStream>(&'a mut self, stream: &mut S) -> KResult<'a, ()> {
        self.finished = stream.read_bits_int(1)? != 0;
        self.reserved = stream.read_bits_int(3)?;
        self.opcode = Some(stream.read_bits_int(4)?.try_into()?);
        self.is_masked = stream.read_bits_int(1)? != 0;
        self.len_payload_primary = stream.read_bits_int(7)?;
        // alignToByte(stream)
        if self.len_payload_primary == 126 {
            self.len_payload_extended_1 = stream.read_u2be()?;
        }
        if self.len_payload_primary == 127 {
            self.len_payload_extended_2 = stream.read_u4be()?;
        }
        if self.is_masked {
            self.mask_key = stream.read_u4be()?;
        }
        if self._root.unwrap().dataframes[0].opcode != Some(WebsocketOpcode::Text) {
            self.payload_bytes = stream.read_bytes(self.len_payload() as usize)?;
        }
        if self._root.unwrap().dataframes[0].opcode == Some(WebsocketOpcode::Text) {
            self.payload_text = panic!("Unimplemented encoding for bytesToStr: {}", "UTF-8");
        }
        Ok(())
    }
}
impl<'a> WebsocketDataframe<'a> {
    fn len_payload(&mut self) -> i32 {
        if let Some(x) = self.len_payload {
            return x;
        }

        // instanceCalculate(InstanceIdentifier(len_payload), CalcIntType, IfExp(Compare(Name(identifier(len_payload_primary)),LtE,IntNum(125)),Name(identifier(len_payload_primary)),IfExp(Compare(Name(identifier(len_payload_primary)),Eq,IntNum(126)),Name(identifier(len_payload_extended_1)),Name(identifier(len_payload_extended_2)))))
        self.len_payload.unwrap()
    }
}
```

Now, besides needing to fill in a couple of last methods (`alignToByte`, conditional repeats), there's one last substantive borrow checker issue:

```
error[E0495]: cannot infer an appropriate lifetime for autoref due to conflicting requirements
   --> src/websocket.rs:131:41
    |
131 |             self.payload_bytes = stream.read_bytes(self.len_payload() as usize)?;
    |                                         ^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined on the method body at 114:5...
   --> src/websocket.rs:114:5
    |
114 | /     fn read<S: KStream>(&'a mut self, stream: &mut S) -> KResult<'a, ()> {
115 | |         self.finished = stream.read_bits_int(1)? != 0;
116 | |         self.reserved = stream.read_bits_int(3)?;
117 | |         self.opcode = Some(stream.read_bits_int(4)?.try_into()?);
...   |
136 | |         Ok(())
137 | |     }
    | |_____^
note: ...so that reference does not outlive borrowed content
   --> src/websocket.rs:131:34
    |
131 |             self.payload_bytes = stream.read_bytes(self.len_payload() as usize)?;
    |                                  ^^^^^^
note: but, the lifetime must be valid for the lifetime 'a as defined on the impl at 102:6...
   --> src/websocket.rs:102:6
    |
102 | impl<'a> KStruct<'a> for WebsocketDataframe<'a> {
    |      ^^
note: ...so that reference does not outlive borrowed content
   --> src/websocket.rs:131:34
    |
131 |             self.payload_bytes = stream.read_bytes(self.len_payload() as usize)?;
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0495`.
error: Could not compile `kaitai_ws`.

To learn more, run the command again with --verbose.
```

I'm having issues with the struct and stream lifetimes tying themselves up. Not sure if @CWood1 has ideas, but I'm planning to study up on some `nom` stuff in the very near future.

bspeice:
Been playing around for a while, here's a minimal replicating situation for the issue I'm having with the borrow checker. If we can get this working, the rest of it should be reasonably easy.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=935b00d4a84392f91ed0d9a9f0d12d1f
```rust
use std::io;

pub trait KStruct<'a> {
    type Parent: KStruct<'a>;
    type Root: KStruct<'a>;

    /// Create a new instance of this struct; if we are the root node,
    /// then both `_parent` and `_root` will be `None`.
    // `_root` is an Option because we need to create a root in the first place
    fn new(_parent: Option<&'a Self::Parent>, _root: Option<&'a Self::Root>) -> Result<Self, ()>
    where
        Self: Sized;

    /// Parse this struct (and any children) from the supplied stream
    fn read<S: KStream>(&'a mut self, stream: &mut S) -> io::Result<()>;
}

pub trait KStream {
    fn read_bytes(&mut self, len: usize) -> io::Result<&[u8]>;
}

struct BytesReader {
    bytes: Vec<u8>,
    pos: usize,
}
impl BytesReader {
    fn new(bytes: Vec<u8>) -> Self {
        BytesReader { bytes, pos: 0 }
    }
}
impl KStream for BytesReader {
    fn read_bytes(&mut self, len: usize) -> io::Result<&[u8]> {
        if len + self.pos > self.bytes.len() {
            return Err(io::Error::from(io::ErrorKind::UnexpectedEof));
        }
        
        Ok(&self.bytes[self.pos..self.pos+len])
    }
}

#[derive(Default)]
struct MyStruct<'a> {
    segment1: &'a [u8],
    segment2: &'a [u8],
    _parent: Option<&'a MyStruct<'a>>,
    _root: Option<&'a MyStruct<'a>>,
}
impl<'a> KStruct<'a> for MyStruct<'a> {
    type Parent = MyStruct<'a>;
    type Root = MyStruct<'a>;
    
    fn new(_parent: Option<&'a Self::Root>, _root: Option<&'a Self::Root>) -> Result<Self, ()> {
        Ok(MyStruct {
            _parent,
            _root,
            ..Default::default()
        })
    }
    
    fn read<S: KStream>(&mut self, stream: &mut S) -> io::Result<()> {
        self.segment1 = stream.read_bytes(4)?;
        self.segment2 = stream.read_bytes(4)?;
        
        Ok(())
    }
}
```

bspeice:
Alright, got a very reduced case working. I can't totally explain it (because I don't totally understand it), but the issue was in the `read` definition, and specifically using `&mut self`. When this happens, the borrow checker is totally unable to prove that this struct doesn't outlive the stream it comes from, and fails out. Even when creating a new instance inside the `read` method and returning `Result<Self>`, there's an implicit lifetime associated with `Self` that causes issues.

Taking some inspiration from `nom`, here's the definition for `KStruct` that I've found allows me to pass the borrow checker. The important bit is that `read` is parameterized on the struct, rather than any references to `Self`.

```rust
pub trait KStruct<'a> {
    type Parent: KStruct<'a>;
    type Root: KStruct<'a>;
    
    fn new(
        parent: Option<&'a Self::Parent>,
        root: Option<&'a Self::Root>
    ) -> Result<Self, ()>
    where
        Self: Sized;
        
    fn read<S: KStruct<'a>>(
        self_: S,
        bytes: &[u8],
    ) -> Result<(&[u8], S), ()>;
}
```

bspeice:
In my ongoing "thinking out loud" mode, here's an example that's able to pass the borrow checker and do some trivial work. I need to clean things up and experiment a bit (I'd rather use `KStream` than raw bytes, etc.), but this is something worth working from:

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a41645490686c3413295abaf1c9c699a)
```rust
pub trait KStruct<'a> {
    type Parent: KStruct<'a> + Sized;
    type Root: KStruct<'a> + Sized;
    
    fn new(parent: Option<&'a Self::Parent>, root: Option<&'a Self::Root>) -> Self
    where
        Self: Sized;
        
    fn parent(&self) -> Option<&Self::Parent>;
    fn root(&self) -> Option<&Self::Root>;
}

#[derive(Default)]
struct MyStruct<'a> {
    pub segment1: &'a [u8],
    parent: Option<&'a MyStruct<'a>>,
    root: Option<&'a MyStruct<'a>>,
}
impl<'a> KStruct<'a> for MyStruct<'a> {
    type Parent = MyStruct<'a>;
    type Root = MyStruct<'a>;
    
    fn new(parent: Option<&'a Self::Parent>, root: Option<&'a Self::Root>) -> Self {
        MyStruct {
            parent,
            root,
            ..Default::default()
        }
    }
    
    fn parent(&self) -> Option<&Self::Parent> {
        self.parent
    }
    
    fn root(&self) -> Option<&Self::Root> {
        self.root.or(Some(self))
    }
}
impl<'a> MyStruct<'a> {
    fn read<'b>(mut self_: MyStruct<'b>, bytes: &'b [u8]) -> Result<(&'b [u8], MyStruct<'b>), ()> {
        let rem = bytes.clone();
        
        // Example parsing out a single byte slice
        let res = rem.split_at(1);
        self_.segment1 = res.0;
        
        Ok((res.1, self_))
    }
}

fn main() {
    let bytes = [1, 2, 3, 4, 5];
    
    // Prove that we can give references away
    let grandparent = MyStruct::new(None, None);
    let parent = MyStruct::new(Some(&grandparent), grandparent.root());
    let child = MyStruct::new(Some(&parent), parent.root());
    
    // And that we can actually read a basic stream
    assert!(MyStruct::read(child, &bytes).is_ok());
}
```

bspeice:
Alright, I've almost got everything working for parsing websocket messages as a test case. The last issue is that all types within the `len_payload` calculation must be the same. Right now, Rust has an issue with this:

```rust
self.len_payload = Some(if self.len_payload_primary <= 125 { self.len_payload_primary } else { if self.len_payload_primary == 126 { self.len_payload_extended_1 } else { self.len_payload_extended_2}} as i32);
```

Specifically, because `len_payload_primary`, `len_payload_extended_1`, and `len_payload_extended_2` are all different types, and Rust doesn't auto-widen, it has no idea what to do with the `if` expression.

@CWood1 - any ideas? After this is done, we can at least merge something that works on websocket messages.

tatref:
Nice work!
About your problem, you can either create an enum, having different variant for each type.
But if all your types are integers, I suppose you could just use an `as` cast? What should be the type of `self.len_payload`?

bspeice:
Fair. There's probably a way to store some context in the expression translator that informs it to cast `as` the correct types. I'll see if there's a simpler solution, but that may be the route this goes down.

bspeice:
Using the `as` expressions in the translator works well! I'm currently running into issues that all languages are having (see [this issue](https://github.com/kaitai-io/kaitai_struct/issues/575)) but otherwise the simple cases are working. I tried compiling some other things today, and there are issues, but this is looking better and better.

bspeice:
Can confirm that I now have a working test for parsing websockets messages! Plenty of work yet to be done, but this is a huge step forward.

bspeice:
Working through everything needed to get this running in CI now. There are a couple issues outstanding:

1. Instance calculations need the stream passed in.
2. Switchable endian-ness
3. Import resolution
4. Specs without a `seq` block have issues because we don't properly close the `impl KStruct` definition (`js_signed_right_shift` and `str_literals` are the problems)

bspeice:
Alright, number 1, 3, and 4 are finished. Switchable endian-ness is going to be a pain for some time, and I'm now dealing with `SwitchType` causing type problems all over the place. From what I can tell, C++ and Java deal with this by letting everything be virtual over `KStruct` when it can't figure out what to do. While Rust can do this as well (by changing the definition of  `KStruct::read()` to take `Option<&KStruct>` instead of `Option<&Self::Root>`), it's far from ideal because we throw away all type information and leave users to do the casting themselves.

I'm looking at how best to resolve `SwitchType` issues in the code, though I expect that adding another enum to handle each of the `cases` variations will be the way to go, rather than just using pointers over `KStruct` like C++/Java.

bspeice:
Alright, did a lot more studying and finally some implementation. C#/Java do in fact allow everything to be virtual, which means that when `SwitchType.combinedType == AnyType`, they're still OK because everything is a boxed pointer. C++/Rust don't have that degree of flexibility, so C++ just blows up.

In Rust, `enum` is designed to handle exactly this case, by being an algebraic sum type. I've finished up most of the code necessary to generate those enums and switch between them during parsing; there's one last issue because I'm having difficulty figuring out what switch variant we're actually assigning to, but the design is essentially done. Thanks to @GreyCat for input on both other languages and Rust.

bspeice:
Good news first - lots of bug fixes.

Bad news - hit an edge case that has implications on the API design. Right now, we pass in `_parent` as a parameter to the `read()` function to help Rust with the lifetimes, which works up until you need [`_parent._parent`](https://github.com/kaitai-io/kaitai_struct_tests/blob/474db78d4bba4692e381f038a3b2a95bb1905c74/formats/nav_parent.ksy). Because Rust never stores that reference anywhere, we have no way of following the chain up.

I'm playing around with some ways to solve this; passing `_parent` references via a stack-like structure won't work well, because the elements within the stack don't have a consistent type. I think I might be able to store a reference just for the lifetime of the function (so children can look up `_parent._parent` only during calls to `read()`), but I'll have more info as this progresses.

bspeice:
Continuing updates! The parent stack idea works well; the API will need some usability tweaks as we go, but works well enough. Now, the interesting bit is that preserving the `_parent._parent...` chain makes the lifetimes a good deal harder. The solution is to split the `KStruct` definition into a "read" and "stream" lifetime; `_root` and `_parent` only need to live for the duration of the `read()` function call, so we mark them as such. After that, the data obtained during `read()` is stored in the actual structs, and is marked as `'stream`. The implementation is available [here](https://github.com/bspeice/kaitai_rust/commit/37ea0f5cc1169163b295e48f409f4bc68e084472).

Ultimately, the split allows us to make sure that children can refer to parents only during the call to `read()`. There's an example available [here](https://github.com/bspeice/kaitai_rust/blob/master/tests/grandparent_validation.rs).

This will require some changes to code generation, but I don't think there will be any more lifetimes issues.

bspeice:
Alright, codegen changes were too much of an issue, tests were breaking all over the place, so I've decided to scrap this branch and start over. Rust work will continue onward, but I'm going to open a new issue to track the implementation.

PR #165, opened 2019-06-04
Added Typescript as a Compiler Target

fudgepop01:
The output of the new "TypeScript" target is typescript that uses a class based format instead of a prototype-based one.
TypeScript allows for autocompletion, making it significantly easier to work with the output of complex formats.
Certain shortcuts were also taken for the time being, namely using a type of `any` instead of the `type 1 | type 2` for switch types

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/TypeScriptClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/TypeScriptClassCompiler.scala
@@ -1,0 +1,76 @@
+package io.kaitai.struct
+
+import io.kaitai.struct.datatype.DataType.{KaitaiStreamType, UserTypeInstream, CalcUserType}
+import io.kaitai.struct.datatype.{Endianness, FixedEndian, InheritedEndian}
+import io.kaitai.struct.format._
+import io.kaitai.struct.languages.TypeScriptCompiler
+import io.kaitai.struct.languages.components.ExtraAttrs
+
+class TypeScriptClassCompiler(
+  classSpecs: ClassSpecs,
+  override val topClass: ClassSpec,
+  config: RuntimeConfig
+) extends ClassCompiler(classSpecs, topClass, config, TypeScriptCompiler) {
+
+  override def compileClass(curClass: ClassSpec): Unit = {
+    provider.nowClass = curClass
+
+    // Forward declarations for recursive types
+    curClass.types.foreach { case (typeName, _) => lang.classForwardDeclaration(List(typeName)) }
+
+    // documentation
+    compileClassDoc(curClass)
+
+    // class thing {...
+    lang.classHeader(curClass.name)
+
+    // static <enum name> = Object.freeze({...})
+    compileEnums(curClass)
+
+    // static <subclass name> = class {...
+    compileSubclasses(curClass)
+
+    provider.nowClass = curClass
+
+
+    // public <attrname>: <attrtype>;
+    val allAttrs: List[MemberSpec] =
+      curClass.seq ++
+      curClass.params ++
+      List(
+        AttrSpec(List(), RootIdentifier, CalcUserType(topClassName, None)),
+        AttrSpec(List(), ParentIdentifier, curClass.parentType)
+      ) ++
+      ExtraAttrs.forClassSpec(curClass, lang)
+    compileAttrReaders(allAttrs)
+
+    // constructor() {...}
+    compileConstructor(curClass)
+
+    compileEagerRead(curClass.seq, curClass.meta.endian)
+
+    // private <private instance name>: ...;
+    // get <public instance name>() {...}
+    compileInstances(curClass)
+
+    // }
+    lang.classFooter(curClass.name)
+  }
+
+  override def compileSeqProc(seq: List[AttrSpec], defEndian: Option[FixedEndian]) = {
+    lang.readHeader(defEndian, seq.isEmpty)
+    compileSeq(seq, defEndian)
+    lang.readFooter()
+  }
+
+  override def compileSeq(seq: List[AttrSpec], defEndian: Option[FixedEndian]) = {
+    var wasUnaligned = false
+    seq.foreach { (attr) =>
+      val nowUnaligned = isUnalignedBits(attr.dataType)
+      if (wasUnaligned && !nowUnaligned)
+        lang.alignToByte(lang.normalIO)
+      lang.attrParse(attr, attr.id, defEndian)
+      wasUnaligned = nowUnaligned
+    }
+  }
+}
--- a/shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
@@ -1,0 +1,722 @@
+package io.kaitai.struct.languages
+
+import io.kaitai.struct.datatype.DataType._
+import io.kaitai.struct.datatype.{Endianness, DataType, FixedEndian, InheritedEndian}
+import io.kaitai.struct.exprlang.Ast
+import io.kaitai.struct.exprlang.Ast.expr
+import io.kaitai.struct.format._
+import io.kaitai.struct.languages.components._
+import io.kaitai.struct.translators.TypeScriptTranslator
+import io.kaitai.struct.{ClassTypeProvider, RuntimeConfig, Utils}
+
+class TypeScriptCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
+  extends LanguageCompiler(typeProvider, config)
+    with ObjectOrientedLanguage
+    with UpperCamelCaseClasses
+    with SingleOutputFile
+    with UniversalDoc
+    with AllocateIOLocalVar
+    with EveryReadIsExpression
+    with FixedContentsUsingArrayByteLiteral {
+  import TypeScriptCompiler._
+
+  override val translator = new TypeScriptTranslator(typeProvider)
+
+  override def indent: String = "  "
+  override def outFileName(topClassName: String): String = s"${type2class(topClassName)}.ts"
+
+  override def outImports(topClass: ClassSpec) = {
+    val impList = importList.toList
+    val quotedImpList = impList.map((x) => s"'$x'")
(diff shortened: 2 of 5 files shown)


shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
```diff
@@ -0,0 +1,722 @@
...
+  override def indent: String = "  "
+  override def outFileName(topClassName: String): String = s"${type2class(topClassName)}.ts"
+
+  override def outImports(topClass: ClassSpec) = {
+    val impList = importList.toList
+    val quotedImpList = impList.map((x) => s"'$x'")
+    val defineArgs = quotedImpList.mkString(", ")
+    val moduleArgs = quotedImpList.map((x) => s"require($x)").mkString(", ")
+    val argClasses = impList.map((x) => x.split('/').last)
+    val rootArgs = argClasses.map((x) => s"root.$x").mkString(", ")
```
GreyCat: Um, pardon my misunderstanding, but it looks like all these values are just being thrown away after calculation?

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
```diff
@@ -0,0 +1,722 @@
...
+    outHeader.puts("interface IDebug {")
+    outHeader.inc
+    outHeader.puts("start?: number;")
+    outHeader.puts("ioOffset?: number;")
+    outHeader.puts("end?: number;")
+    outHeader.puts("arr?: IDebug[];")
+    outHeader.puts("enumName?: string;")
+    outHeader.puts("[key: string]: number | string | IDebug | IDebug[] | undefined;")
+    outHeader.dec
+    outHeader.puts("}")
```
GreyCat: It looks like this warrants to place it in runtime, not repeat it in every generated file?

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
fudgepop01: oh! Those are leftover from the JavaScript compiler. I used that as a base and modified it heavily. Yeah, those can be deleted if i'm not mistaken

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
```diff
@@ -0,0 +1,722 @@
...
+  def types2class(types: List[String]): String = types.map(type2class).mkString(".")
+
+  override def type2class(name: String): String = Utils.upperCamelCase(name)
+
+  def isUserOrEnumType(inType: DataType): String = inType match {
+    case _: UserType => "user"
+    case _: EnumType => "enum"
+    case ArrayType(innerType) => isUserOrEnumType(innerType)
+    case _ => "other"
+  }
```
GreyCat: Um, no. You don't really need any strings in there. If you really want this, you can perfectly return `DataType` and match with these `case _: UserType` in every place where you match with `case "user"`, etc. OTOH, I'm not even sure why this is needed.

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
```diff
@@ -0,0 +1,722 @@
...
+      case KaitaiStructType | CalcKaitaiStructType => "any"
+
+      case st: SwitchType => kaitaiType2NativeType(st.combinedType, absolute)
+    }
+  }
+
+  override def kstreamName: String = "KaitaiStream"
+
+  // FIXME: probably KaitaiStruct will emerge some day in JavaScript runtime, but for now it is unused
+  override def kstructName: String = ???
```
GreyCat: Actually, it's a very good question. Probably TypeScript can make this interface a reality?

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
fudgepop01: that might be a good idea, but the typescript runtime doesn't exist yet. Until it does, it's sadly necessary to include it in at least debug mode compilation

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
fudgepop01: Apologies, i'm not too familiar with scala (this is literally the only experience i've had with it so-far). I can definitely work on cleaning stuff up- I just needed to get this reviewed to understand what should be done and where

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
fudgepop01: This will likely happen once a runtime for typescript is created. I wasn't quite sure whether or not I should go ahead and name it so I just left it as-is

shared/src/main/scala/io/kaitai/struct/TypeScriptClassCompiler.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    // }
+    lang.classFooter(curClass.name)
+  }
+
+  override def compileSeqProc(seq: List[AttrSpec], defEndian: Option[FixedEndian]) = {
+    lang.readHeader(defEndian, seq.isEmpty)
+    compileSeq(seq, defEndian)
+    lang.readFooter()
+  }
```
GreyCat: `compileSeqProc` seems to be 100% equal of original implementation in ClassCompiler. Why would you need this override + copy/paste?

shared/src/main/scala/io/kaitai/struct/TypeScriptClassCompiler.scala
```diff
@@ -0,0 +1,76 @@
...
+  override def compileSeq(seq: List[AttrSpec], defEndian: Option[FixedEndian]) = {
+    var wasUnaligned = false
+    seq.foreach { (attr) =>
+      val nowUnaligned = isUnalignedBits(attr.dataType)
+      if (wasUnaligned && !nowUnaligned)
+        lang.alignToByte(lang.normalIO)
+      lang.attrParse(attr, attr.id, defEndian)
+      wasUnaligned = nowUnaligned
+    }
+  }
```
GreyCat: Again, `compileSeq` seems to be full equivalent of ClassCompiler implementation. Why do we need that?

fudgepop01: you're right on both accounts. I put them there initially because I thought that I would modify them somehow, however it turns out that's not the case.

shared/src/main/scala/io/kaitai/struct/translators/TypeScriptTranslator.scala
```diff
@@ -0,0 +1,135 @@
...
+  override def arraySize(a: expr): String =
+    s"${translate(a)}.length"
+  override def arrayMin(a: expr): String =
+    s"${TypeScriptCompiler.kstreamName}.arrayMin(${translate(a)})"
+  override def arrayMax(a: expr): String =
+    s"${TypeScriptCompiler.kstreamName}.arrayMax(${translate(a)})"
+
+  override def kaitaiStreamEof(value: Ast.expr): String =
+    s"${translate(value)}.isEof()"
+}
```
GreyCat: Missing newline.

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
```diff
@@ -0,0 +1,722 @@
...
+      case IntMultiType(false, Width4, _) => "number"
+      case IntMultiType(false, Width8, _) => "number"
+
+      case Int1Type(true) => "number"
+      case IntMultiType(true, Width2, _) => "number"
+      case IntMultiType(true, Width4, _) => "number"
+      case IntMultiType(true, Width8, _) => "number"
+
+      case FloatMultiType(Width4, _) => "number"
+      case FloatMultiType(Width8, _) => "number"
```
GreyCat: All these can be probably reduced to:

```scala
case _: NumericType => "number"
```

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
```diff
@@ -0,0 +1,722 @@
...
+    id match {
+      case SpecialIdentifier(name) => name
+      case NamedIdentifier(name) => Utils.lowerCamelCase(name)
+      case NumberedIdentifier(idx) => s"_${NumberedIdentifier.TEMPLATE}$idx"
+      case InstanceIdentifier(name) => s"_m_${Utils.lowerCamelCase(name)}"
+      case RawIdentifier(innerId) => "_raw_" + idToStr(innerId)
+    }
+  }
+
+  override def privateMemberName(id: Identifier): String = s"${idToStr(id)}"
```
GreyCat: I see that you've removed `this.` from `privateMemberName`, and instead appended it everywhere into the generated statements themselves. What is a reason for that? Normally, in most codegens, `privateMemberName` vs `idToStr` indeed is used to do `this.foo`, `self.foo`, `@foo`, etc, i.e. internal private member names out of an `Identifier`.

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
GreyCat: Um, pardon me if I'm wrong, but from what I see so far, forking and implementing a TS runtime is 10x easier than what you've achieved already — and it will save from many existing kludges in JS runtime? Why not take that extra step — given that it's literally creation of 1 new repo with 1-2 files and that's it?

shared/src/main/scala/io/kaitai/struct/languages/TypeScriptCompiler.scala
fudgepop01: when I did that before, the output caused the declarations to also contain "this.", causing errors. I'm not aware of how to filter that out in that one specific instance

PR #166, opened 2019-06-05
Custom Scan enhancement #538

tinrodriguez8:
Adding first things for custom scan and simulating terminator byte.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/ConstructClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ConstructClassCompiler.scala
@@ -136,7 +136,7 @@ class ConstructClassCompiler(classSpecs: ClassSpecs, topClass: ClassSpec) extend
       s"Int${width.width * 8}${signToStr(signed)}${fixedEndianToStr(endianOpt.get)}"
     case FloatMultiType(width, endianOpt) =>
       s"Float${width.width * 8}${fixedEndianToStr(endianOpt.get)}"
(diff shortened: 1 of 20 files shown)


tinrodriguez8:
Talking with @GreyCat he noticed me some things:

> I've taken a look, so on a broader scale:
I won't touch BytesEosType and BytesTerminatedType — I'd say that there's very little value in having both size + scan-end support at once, or terminator + scan-end
So, BytesScanEndType alone would probably more than enough.

Should I extend `BytesType` with `ScanEnd`? or maybe it would be better do something like this:
```
case class BytesScanEndType(
    override val process: Option[ProcessExpr],
    override val scanEnd: Option[ScanExpr] = None
  ) extends BytesType with ScanEnd
```

> Also, for ScanExpr, it's probably just enough to have ScanCustom arguments right in the ScanExpr, and avoid doing that "trait inherited by multiple classes" thing, and then switching it over every time like this: https://github.com/kaitai-io/kaitai_struct_compiler/compare/538#diff-121a1b8f73b9f24552fa03a24f589eedR191
you can basically remove ScanExpr, replace every time it's mentioned with ScanCustom, and rename ScanCustom to ScanExpr

GreyCat:
> ```scala
> case class BytesScanEndType(
>     override val process: Option[ProcessExpr],
>     override val scanEnd: Option[ScanExpr] = None
>   ) extends BytesType with ScanEnd
> ```

Yep, this looks like the best solution: there's no need to add that functionality to every bytes type, a dedicated feature to will be the best.

PR #167, opened 2019-08-01
Update Scala.js to 0.6.28

flucknugget:
Adds ES6 module support, see https://github.com/scala-js/scala-js/commit/4fffdf61e23916fe17630e826ce55325b9bf6c0e


flucknugget:
I tested this locally and everything seems to work.

GreyCat:
Makes sense, thanks!

PR #168, opened 2019-09-09
Support some instances of process in Go.

jchv:
As-is, the Go runtime already contains functionality to support much of `process`, but today the compiler doesn't emit code to use it. This PR attempts to add support for many basic uses of `process`. Custom will remain a noop, as it is currently, because otherwise the unit tests fail to run at all.

Before: `SUMMARY: {"kst"=>77, "passed"=>61, "failed"=>24}`
After: `SUMMARY: {"kst"=>77, "passed"=>71, "failed"=>14}`

(This is my first PR for anything written in Scala. If something looks stupid, it probably is.)


GreyCat:
Everything looks cool, thanks :) Merging in, will hope to kick off the CI builds tomorrow, when I'll get back to my desktop.

PR #169, opened 2019-09-10
Improve switch statement support in Go

jchv:
Another round of Go improvements. This time, various kinds of switch statements now work. None of the kst tests seem to rely on switch statements, so this time there is no change in tests passing or failing. However, more structs are successfully compiled.

Before:
`SUMMARY: {"kst"=>77, "passed"=>71, "failed"=>14}`

After:
`SUMMARY: {"kst"=>77, "passed"=>71, "failed"=>14}`

##### Switched type
```go
	this.Code = tmp1
	switch (this.Code) {
	case 1:
		tmp2, err := this._io.ReadU1()
		if err != nil {
			return err
		}
		this.Len = uint64(tmp2)
	case 2:
		tmp3, err := this._io.ReadU2le()
		if err != nil {
			return err
		}
		this.Len = uint64(tmp3)
...
```

##### Byte-slice switch
```go
	switch (true) {
	case bytes.Equal(this.Code, []uint8{73}):
		tmp4 := new(SwitchBytearray_Opcode_Intval)
		err = tmp4.Read(this._io, this, this._root)
		if err != nil {
			return err
		}
		this.Body = tmp4
	case bytes.Equal(this.Code, []uint8{83}):
		tmp5 := new(SwitchBytearray_Opcode_Strval)
		err = tmp5.Read(this._io, this, this._root)
		if err != nil {
			return err
		}
		this.Body = tmp5
	}
```

(Note: The `switch (true) {` can just be `switch {`, but to keep it simple I did not bother implementing this simplification. This code is semantically identical.)

#### Newly-working compilations
It's a little hard to quantify how this improves things since kst tests don't cover it. Instead, here is a diff of files that both compile and survive the recovery phase of unit testing. So, these are files that either failed to compile before, or contained errors, that now compile in this branch.

```diff
+nav_parent2.go
+nav_parent3.go
+nav_parent_switch.go
+non_standard.go
+process_coerce_switch.go
+recursive_one.go
+switch_bytearray.go
+switch_integers2.go
+switch_integers.go
+switch_manual_enum.go
+switch_manual_enum_invalid_else.go
+switch_manual_enum_invalid.go
+switch_manual_int_else.go
+switch_manual_int.go
+switch_manual_int_size_else.go
+switch_manual_int_size_eos.go
+switch_manual_int_size.go
+switch_manual_str_else.go
+switch_manual_str.go
+switch_multi_bool_ops.go
+switch_repeat_expr.go
```

It is a little frustrating that the Go compiler duplicates some functionality, but it doesn't seem like most of these customizations will be useful to other languages, and the Go compiler already deviates unusually far from the norm. So this trend is likely to continue, and maybe could be alleviated some day by trying to build better abstractions.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
@@ -318,22 +318,21 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
 //    expr2
 //  }
 
-  override def switchStart(id: Identifier, on: Ast.expr): Unit =
+  override def switchStart(id: Identifier, on: Ast.expr): Unit = {
     out.puts(s"switch (${expression(on)}) {")
+  }
 
   override def switchCaseStart(condition: Ast.expr): Unit = {
-    out.puts(s"case ${expression(condition)}: {")
+    out.puts(s"case ${expression(condition)}:")
     out.inc
   }
 
   override def switchCaseEnd(): Unit = {
-    out.puts("break;")
     out.dec
-    out.puts("}")
   }
 
   override def switchElseStart(): Unit = {
-    out.puts("default: {")
+    out.puts("default:")
     out.inc
   }
 
--- a/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
@@ -5,7 +5,7 @@ import io.kaitai.struct.datatype.{DataType, FixedEndian, KSError}
 import io.kaitai.struct.exprlang.Ast
 import io.kaitai.struct.format._
 import io.kaitai.struct.languages.components._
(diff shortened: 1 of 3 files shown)


jchv:
Oh, hmm. I think I just realized what's going on. It looks like there are indeed kst tests for some of these things, but the Go variations are not generated, presumably because they would cause the test suite to fail catastrophically before. Let me see if I can figure out how to generate tests for what is newly working.

jchv:
Wow, looks like the Go kst translator may need to catch up to the state of the Go compiler soon. I was only able to compile a few new working tests. I am happy to report, however, that all of the newly added tests pass.

`SUMMARY: {"kst"=>82, "passed"=>76, "failed"=>14}`

```diff
+ spec/go/nav_parent2_test.go
+ spec/go/nav_parent3_test.go
+ spec/go/non_standard_test.go
+ spec/go/switch_integers2_test.go
+ spec/go/switch_integers_test.go
```

PR to add newly-passing tests: kaitai-io/kaitai_struct_tests#55.

GreyCat:
Thanks for this impressive effort, looks good to me, merging in!

PR #170, opened 2019-09-11
[WIP] Nim Compiler

sealmove:
```
meta:
  id: hello_world
seq:
  - id: one
    type: u1
```

``` nim
# This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

import ../../runtime/nim/kaitai

type
  HelloWorld* = ref object
    one*: uint8
    root*: HelloWorld
    parent*: ref RootObj

proc read*(_: typedesc[HelloWorld], stream: KaitaiStream, root: HelloWorld, parent: ref RootObj): owned HelloWorld =
  result = new(HelloWorld)
  let root = if root == nil: result else: root
  result.one = readU1(stream)
  result.root = root
  result.parent = parent

proc fromFile*(_: typedesc[HelloWorld], filename: string): owned HelloWorld =
  var stream = newKaitaiStream(filename)
  HelloWorld.read(stream, nil, nil)
```

I think this pattern is flexible and neat:
- The user will be able to create objects of any type (subsections of a format) by simply providing the type and a file path (fromFile()). In this case the object returned will be a root object. For example he will be able to create a Png Chunks object like this: `Chunks.fromFile("path/to/partial/file")`
- read() is mostly for internal use, but it's also exposed to the user in case he needs to do something more involved (for example construct a png from partial png files).

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
@@ -17,17 +17,18 @@ class NimCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
   import NimCompiler._
 
   // Members declared in io.kaitai.struct.languages.components.ExceptionNames
-  override def ksErrorName(err: KSError): String = "" // XXX
+  override def ksErrorName(err: KSError): String = ""
 
   // Members declared in io.kaitai.struct.languages.components.ExtraAttrs
-  override def extraAttrForIO(id: Identifier, rep: RepeatSpec): List[AttrSpec] = List() // XXX
+  override def extraAttrForIO(id: Identifier, rep: RepeatSpec): List[AttrSpec] = List()
 
   // Members declared in io.kaitai.struct.languages.components.LanguageCompiler
-  override def alignToByte(io: String): Unit = () // XXX
-  override def attrFixedContentsParse(attrName: Identifier, contents: Array[Byte]): Unit = Array() // XXX
-  override def attrParse(attr: AttrLikeSpec, id: Identifier, defEndian: Option[Endianness]): Unit = Option() // XXX
-  override def attrParseHybrid(leProc: () => Unit, beProc: () => Unit): Unit = () // XXX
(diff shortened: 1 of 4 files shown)


sealmove:
`hello_world.ksy` now compiles correctly!
The basic architecture for Nim compiler is now set up.
Next step is to write all the tests.

shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
```diff
@@ -0,0 +1,206 @@
...
+  override def condRepeatEosHeader(id: Identifier, io: String, dataType: DataType, needRaw: Boolean): Unit = ()
+  override def condRepeatExprFooter: Unit = ()
+  override def condRepeatExprHeader(id: Identifier, io: String, dataType: DataType, needRaw: Boolean, repeatExpr: Ast.expr): Unit = ()
+  override def condRepeatUntilFooter(id: Identifier, io: String,dataType: DataType,needRaw: Boolean, repeatExpr: Ast.expr): Unit = ()
+  override def condRepeatUntilHeader(id: Identifier, io: String,dataType: DataType,needRaw: Boolean, repeatExpr: Ast.expr): Unit = ()
+  override def enumDeclaration(curClass: List[String], enumName: String, enumColl: Seq[(Long, EnumValueSpec)]): Unit = ()
+
+  override def fileHeader(topClassName: String): Unit = {
+    outHeader.puts(s"# $headerComment")
+    importList.add(s"../../../runtime/nim/kaitai")
```
GreyCat: I guess this refers to the location of these files in test repo vs runtimes repo, but it won't be always true for production use?

GreyCat:
Looks good to me, thanks for this contribution! Merging in.

shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
sealmove: I noticed `RuntimeConfig.scala` and the relevant command line options (which exist for only ~half of the supported languages) but still don't understand the approach here. Is it safe to guess that it's a matter of having a saner default and adding a command line option for specifying the module's path? If so, what should this default be (relevant to the path of the user's project)?

PR #171, opened 2019-09-23
Add support for endianness-switch to Go compiler.

jchv:
This enables Kaitai Struct files that switch on endian to be compiled with the Go compiler. In addition, it cleans up some incongruencies between Go and other languages.

Newly compiling Kaitai Structs:
```diff
+default_endian_expr_exception.go
+default_endian_expr_inherited.go
+default_endian_expr_is_be.go
+default_endian_expr_is_le.go
```

Test summary (same before/after, need to compile+add more tests):
```
SUMMARY: {"kst"=>82, "passed"=>76, "failed"=>14}
```

(Will look into running more tests soon. It's getting late here.)

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
@@ -1,11 +1,11 @@
 package io.kaitai.struct.languages
 
 import io.kaitai.struct.datatype.DataType._
-import io.kaitai.struct.datatype.{DataType, FixedEndian, KSError}
+import io.kaitai.struct.datatype.{CalcEndian, DataType, FixedEndian, InheritedEndian, KSError}
 import io.kaitai.struct.exprlang.Ast
 import io.kaitai.struct.format._
 import io.kaitai.struct.languages.components._
-import io.kaitai.struct.translators.{GoTranslator, ResultString, TranslatorResult, TypeDetector}
+import io.kaitai.struct.translators.{GoTranslator, ResultString, TranslatorResult}
 import io.kaitai.struct.{ClassTypeProvider, RuntimeConfig, Utils}
 
 class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
@@ -32,7 +32,7 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
   override def outFileName(topClassName: String): String =
     s"src/${config.goPackage}/$topClassName.go"
 
-  override def outImports(topClass: ClassSpec) = {
+  override def outImports(topClass: ClassSpec): String = {
     val imp = importList.toList
     imp.size match {
       case 0 => ""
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
```diff
@@ -62,33 +62,73 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     out.inc
   }
 
-  override def classFooter(name: List[String]): Unit = universalFooter
+  override def classFooter(name: List[String]): Unit = {
+    // TODO(jchw): where should this attribute actually be generated at?
```
GreyCat: Um, what do you mean by "this attribute"?

shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
```diff
@@ -62,33 +62,73 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
+    out.dec
+    out.puts("case 1:")
+    out.inc
+    out.puts("err = this._read_le()")
+    out.dec
+    out.puts("default:")
+    out.inc
+    // FIXME
+    // out.puts(s"err = ${GoCompiler.ksErrorName(UndecidedEndiannessError)}()")
+    out.puts("panic(\"undecided endianness\")")
```
GreyCat: This actually facilitates a pretty untrivial challenge: ideally, we'd like to stick to Go way here and avoid doing any kinds of exceptions / panics here, and stick with return codes. But this would probably mean that `Read` will change dramatically?

GreyCat:
Thanks, this is indeed a pretty impressive body of work!

shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
jchv: I think we’re fine: read does return an error value. The fixme is because there’s no undecided endian error in the Go runtime.

shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
jchv: Whoops, i just mean the is_le struct field.

shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
GreyCat: Sounds good, it should be easy to fix eventually then :)

shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
GreyCat: Ok, what do you mean by "generation" of is_le attribute then? Declaration of it in a struct?

GreyCat:
CI seems to have finished, looks ok. Can you go ahead and add the new tests you've mentioned?

> default_endian_expr_exception.go
> default_endian_expr_inherited.go
> default_endian_expr_is_be.go
> default_endian_expr_is_le.go

jchv:
Of course, but I’ll have to look after work. We’re getting close to the point of needing to improve the kst compiler for Go I believe.

PR #174, opened 2019-09-26
Nim Compiler

sealmove:
This PR goes together with [this one](https://github.com/kaitai-io/kaitai_struct_tests/pull/60).

Another big phase of codegen fixes and a draft version of NimTranslator.
These are just enough so that the test suite can produce results.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
@@ -8,7 +8,7 @@ import io.kaitai.struct.exprlang.Ast.expr
 import io.kaitai.struct.format._
 import io.kaitai.struct.languages.components._
 
-import io.kaitai.struct.translators.{CppTranslator, TypeDetector}
+import io.kaitai.struct.translators.{NimTranslator, TypeDetector}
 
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
```diff
@@ -100,7 +104,7 @@ class NimCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
   override def seek(io: String, pos: Ast.expr): Unit = ()
 
   val importListSrc = new ImportList
-  override val translator: io.kaitai.struct.translators.AbstractTranslator = new CppTranslator(typeProvider, importListSrc, config)
+  override val translator: io.kaitai.struct.translators.AbstractTranslator = new NimTranslator(typeProvider, importListSrc)
```
GreyCat: Likely just `: AbstractTranslator` should work?

GreyCat:
Thank you, merging this one in as well!

shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
sealmove: Thanks, missed this one

PR #178, opened 2019-11-09
Fix npm publish

cherue:
Currently no new versions are published to npm because npm versions must follow semver and Travis didn't catch the error. This pull request fixes both these issues.


GreyCat:
Let's go with it, thanks for this contribution!

PR #179, opened 2019-11-15
Fix repetitions in 'serialization', add repeat-until last element check

generalmimon:
See https://github.com/kaitai-io/kaitai_struct/issues/27

I realized that we don't need **special loops** for each repetition type, because in case of writing, we iterate some array of known size and write them consecutively in the order they appear in the array, so deriving the number of repetitions (repeat-expr) or checking a condition for each element (repeat-until) isn't necessary.

I added a check if the last value of array satisfies the condition of **repeat-until** to ensure format consistency.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/exprlang/Ast.scala
+++ b/shared/src/main/scala/io/kaitai/struct/exprlang/Ast.scala
@@ -106,6 +106,8 @@ object Ast {
     case class Subscript(value: expr, idx: expr) extends expr
     case class Name(id: identifier) extends expr
     case class List(elts: Seq[expr]) extends expr
+
+    case class CodeLiteral(code: String) extends expr
   }
 
   sealed trait boolop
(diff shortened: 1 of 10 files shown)


GreyCat:
This PR seems to bundle about a dozen of changes together, and most of them are not explained neither in PR comment, nor in commit notes.

Can I ask you to either split it into smaller pieces, or, if they're all tightly coupled, we'll need more detailed explanations on what and why has changed there.

GreyCat:
If the second alternative seems to be better, I can start asking questions like "what is it", if that would be of any help?

generalmimon:
> This PR seems to bundle about a dozen of changes together, and most of them are not explained neither in PR comment, nor in commit notes.

I apologize, this PR is actually the result of solving a few exceptions which occured when I was running tests while merging master into serialization branch (https://github.com/kaitai-io/kaitai_struct_tests/pull/67). All changes are not tightly coupled, I'll try to split them in smaller pieces.

So I close this PR and create several new ones. Anyway, thank you for your response.

PR #180, opened 2019-11-17
Stop generating empty constructor when autoRead

generalmimon:
If someone would compile KSY format with --read-write option only (without --no-auto-read), instancing any user-generated class with empty parameter list would immediately fail with an exception (because autoRead is enabled, but there is no _io to read from).


GreyCat:
"Autoread" at all makes sense only if we're in read-only mode. Otherwise, it's bad idea to even allow autoread constructors. We can actually hard-wire `--read-write` option to disable autoread, so we won't care about all these exceptional handling around the code?

GreyCat:
Thanks!

PR #182, opened 2019-11-17
Use common loop for all repetitions

generalmimon:
I realized that we don't need **special loops** for each repetition type, because in case of writing, we iterate some array of known size and write them consecutively in the order they appear in the array, so deriving the number of repetitions (repeat-expr) or checking a condition for each element (repeat-until) isn't necessary.


GreyCat:
Totally makes sense, thanks!

PR #184, opened 2019-11-24
Fix byte array operations in PHP

generalmimon:
Solves https://github.com/kaitai-io/kaitai_struct/issues/633 and satisfies test [ExprBytesOps](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/formats/expr_bytes_ops.ksy) for PHP


GreyCat:
Thanks!

PR #185, opened 2019-12-07
Allow _parent call on generic KaitaiStructType

generalmimon:
Solves https://github.com/kaitai-io/kaitai_struct/issues/200


GreyCat:
LGTM, thanks!

PR #186, opened 2019-12-10
Fix enum generation for values with id bigger than Int.MAX_VALUE

valery1707:
This fix allow to compile Java-classes with enum with `id` which value is bigger than `Int.MAX_VALUE`.

This PR is inspired by https://github.com/valery1707/kaitai-gradle-plugin/issues/17.

If `*.ksy` has code with something like
```
enums:
  profile_ids:
    0x00000003: defaults                # Known default value is used.
    0xFFFFFFFF: not_found               # Value not found.
```

in generated Java-class will be generated something like
```java
public enum PROFILE_IDS {
  DEFAULTS(3),
  NOT_FOUND(4294967295);

  private final long id;

  PROFILE_IDS(long id) {
    this.id = id;
  }
}
```

But in this case argument will be interpreted as `Int`, and value `4294967295` is not valid for integer range and on compilation user receive error `integer number too large`.
With this fix ids for enums will be generated in valid long-safe mode:
```java
public enum PROFILE_IDS {
  DEFAULTS(3),
  NOT_FOUND(4294967295L);

  private final long id;

  PROFILE_IDS(long id) {
    this.id = id;
  }
}
```

P.S.
Is this project has some test of generation results?

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
@@ -661,12 +661,12 @@ class JavaCompiler(val typeProvider: ClassTypeProvider, config: RuntimeConfig)
 
     if (enumColl.size > 1) {
       enumColl.dropRight(1).foreach { case (id, label) =>
-        out.puts(s"${value2Const(label)}($id),")
+        out.puts(s"${value2Const(label)}(${long2str(id)}),")
       }
     }
     enumColl.last match {
       case (id, label) =>
-        out.puts(s"${value2Const(label)}($id);")
+        out.puts(s"${value2Const(label)}(${long2str(id)});")
     }
 
     out.puts
@@ -696,6 +696,14 @@ class JavaCompiler(val typeProvider: ClassTypeProvider, config: RuntimeConfig)
   }
 
   def value2Const(s: String) = s.toUpperCase
+
+  def long2str(l: Long): String = {
+    if (l.isValidInt) {
+      l.toString()
+    } else {
+      l.toString() + "L"
+    }
+  }
 
   def idToStr(id: Identifier): String = {
     id match {


GreyCat:
Thanks for this investigation and a fix! As a matter of fact, [we do have extensive test suite](https://github.com/kaitai-io/kaitai_struct_tests), which results in [this CI dashboard](https://ci.kaitai.io/). Can I ask you to do the following?

* Go to compiler repo
  * Run `sbt publishLocal`
* Go to tests repo
  * Create test case that demonstrates the problem — something like "enum_long.ksy" in [formats/](https://github.com/kaitai-io/kaitai_struct_tests/tree/master/formats) dir of that repo
  * Create spec for that test case using KST format (something like "enum_long.kst") in [spec/ks/](https://github.com/kaitai-io/kaitai_struct_tests/tree/master/spec/ks) dir of that repo. This is basically a language-neutral way to generate test specs: this [enum_0.kst](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec/ks/enum_0.kst) would automatically generate [specs for Java](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec/java/src/io/kaitai/struct/spec/TestEnum0.java) and for all other languages.
  * Commit ksy+kst
  * Run `./spec-kst-for-all -f enum_long`
  * Commit the results of spec generation
* Send this all as PR to tests repo

If not, don't worry, I'll do it myself, it will just take slightly longer :)

valery1707:
@GreyCat ok i will do it soon.
Presumably today.

valery1707:
@GreyCat, I created PR for test suite: https://github.com/kaitai-io/kaitai_struct_tests/pull/70

GreyCat:
Got it, it's now going through CI, we'll see how bad that situation is across all languages, and then I'll merge this PR.

PR #187, opened 2019-12-17
Fix C++ generated code causing GCC warnings

maks500:
Some of the code generated by kaitai for C++ was triggering compiler warnings, this should fix the unused parameters, sign conversions and sign compare ones.  

Given that this code calls a new method `kaitai::to_signed` it strictly depends on https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/33 being merged as well, otherwise this won't work.


GreyCat:
Unfortunately, reverting this one too to get C++98 support back.

PR #189, opened 2020-01-03
WIP: C# Async with .NET Core Kaitai.Struct.Runtime

pluskal:
Relates to https://github.com/kaitai-io/kaitai_struct/issues/640

This is a Work In Progress. I expect to rebase to keep this branch up-to-date.


shared/src/main/scala/io/kaitai/struct/translators/CSharpAsyncTranslator.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+  override def bytesToStr(bytesExpr: String, encoding: Ast.expr): String =
+    s"System.Text.Encoding.GetEncoding(${translate(encoding)}).GetString($bytesExpr)"
+  override def strLength(s: expr): String =
+    s"${translate(s)}.Length"
+
+  // FIXME: This is not fully Unicode aware, but might be better than nothing.
+  // http://stackoverflow.com/a/228060/2055163
+  override def strReverse(s: expr): String =
+    s"new string(Array.Reverse(${translate(s)}.ToCharArray()))"
```
generalmimon, line 110: The current C# implementation is Unicode-aware, so it might be useful to use the same implementation - see [`kaitai_struct_csharp_runtime/KaitaiStream.cs:686-700`](https://github.com/kaitai-io/kaitai_struct_csharp_runtime/blob/1b66a9a85f39c52728893400a23504844cc78e34/KaitaiStream.cs#L686-L700). This is how it's called from the compiler currently:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/361ca50a9b1a29aef431636e9643eee550b8354f/shared/src/main/scala/io/kaitai/struct/translators/CSharpTranslator.scala#L109-L110

MeikelLP:
@pluskal are you still working on this? I think this can be a huge improvement but it seems to be stale for 3+ years...

PR #190, opened 2020-01-03
Fix invalid PHP namespaces

generalmimon:
Closes https://github.com/kaitai-io/kaitai_struct/issues/637


GreyCat:
Seems to be legit and solves the problem. Thanks!

PR #191, opened 2020-01-11
Generate annotations for Java classes

Mingun:
Proposal for https://github.com/kaitai-io/kaitai_struct_java_runtime/issues/22. Annotations make it much easier to create a visualizer, especially, if input for visualizer is class, derived from kaitai-generated class and if https://github.com/kaitai-io/kaitai_struct/issues/672 will be implemented.

Right now I want to see, if it compiles and also receive some review comments.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
@@ -44,6 +44,7 @@ class ClassCompiler(
 
     if (!lang.innerDocstrings)
       compileClassDoc(curClass)
+    lang.classAnnotation(curClass)
     lang.classHeader(curClass.name)
     if (lang.innerDocstrings)
       compileClassDoc(curClass)
@@ -90,15 +91,18 @@ class ClassCompiler(
 
     // Attributes declarations and readers
     val allAttrs: List[MemberSpec] =
-      curClass.seq ++
-      curClass.params ++
       List(
         AttrSpec(List(), RootIdentifier, topClass.toDataType),
         AttrSpec(List(), ParentIdentifier, curClass.parentType)
       ) ++
       ExtraAttrs.forClassSpec(curClass, lang)
+
+    compileIndexedAttrReaders(curClass.seq)
+    compileIndexedAttrReaders(curClass.params)
     compileAttrReaders(allAttrs)
 
+    compileIndexedAttrDeclarations(curClass.seq)
+    compileIndexedAttrDeclarations(curClass.params)
     // NOTE: most target languages don't care, but Zig requires all fields to be
     // declared together, see https://github.com/ziglang/zig/issues/5077
     compileInstanceDeclarations(curClass)
@@ -203,12 +207,7 @@ class ClassCompiler(
     */
   def compileAttrDeclarations(attrs: List[MemberSpec]): Unit = {
     attrs.foreach { (attr) =>
-      val isNullable = if (lang.switchBytesOnlyAsRaw) {
-        attr.isNullableSwitchRaw
-      } else {
-        attr.isNullable
-      }
-      lang.attributeDeclaration(attr.id, attr.dataTypeComposite, isNullable)
+      lang.attributeDeclaration(attr.id, attr.dataTypeComposite, isNullable(attr))
     }
   }
 
@@ -222,15 +221,49 @@ class ClassCompiler(
       // FIXME: Python should have some form of attribute docs too
       if (!attr.doc.isEmpty && !lang.innerDocstrings)
         lang.attributeDoc(attr.id, attr.doc)
-      val isNullable = if (lang.switchBytesOnlyAsRaw) {
-        attr.isNullableSwitchRaw
-      } else {
-        attr.isNullable
-      }
-      lang.attributeReader(attr.id, attr.dataTypeComposite, isNullable)
+      val isNull = isNullable(attr)
+      lang.attributeReader(attr.id, attr.dataTypeComposite, isNull)
       if (config.readWrite) {
-        lang.attributeSetter(attr.id, attr.dataTypeComposite, isNullable)
+        lang.attributeSetter(attr.id, attr.dataTypeComposite, isNull)
       }
+    }
+  }
+
+  /**
+    * Iterates over a given list of attributes and generates attribute
+    * declarations for each of them, additionally annotate each field.
+    *
+    * @param attrs attribute list to traverse
+    */
+  def compileIndexedAttrDeclarations(attrs: List[MemberSpec]): Unit = {
+    attrs.zipWithIndex.foreach { case (attr, index) =>
+      lang.attributeAnnotation(index, attr)
+      lang.attributeDeclaration(attr.id, attr.dataTypeComposite, isNullable(attr))
+    }
+  }
+
+  /**
+    * Iterates over a given list of attributes and generates attribute
+    * readers (AKA getters) for each of them, additionally annotate each
+    * getter.
+    *
+    * @param attrs attribute list to traverse
+    */
+  def compileIndexedAttrReaders(attrs: List[MemberSpec]): Unit = {
+    attrs.zipWithIndex.foreach { case (attr, index) =>
+      // FIXME: Python should have some form of attribute docs too
+      if (!attr.doc.isEmpty && !lang.innerDocstrings)
+        lang.attributeDoc(attr.id, attr.doc)
+      lang.attributeAnnotation(index, attr)
+      lang.attributeReader(attr.id, attr.dataTypeComposite, isNullable(attr))
+    }
+  }
+
+  def isNullable(attr: MemberSpec): Boolean = {
+    if (lang.switchBytesOnlyAsRaw) {
+      attr.isNullableSwitchRaw
+    } else {
+      attr.isNullable
     }
   }
 
@@ -424,6 +457,7 @@ class ClassCompiler(
 
     if (!lang.innerDocstrings)
       compileInstanceDoc(instName, instSpec)
+    lang.instanceAnnotation(instName, instSpec, true)
     lang.instanceHeader(className, instName, dataType, instSpec.isNullable)
     if (lang.innerDocstrings)
       compileInstanceDoc(instName, instSpec)
@@ -477,6 +511,7 @@ class ClassCompiler(
     } else {
       instSpec.isNullable
     }
+    lang.instanceAnnotation(instName, instSpec, false)
     lang.instanceDeclaration(instName, instSpec.dataTypeComposite, isNullable)
     if (config.readWrite)
       instSpec match {
--- a/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
@@ -868,8 +868,8 @@ class JavaCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
 
   //</editor-fold>
 
-  override def instanceDeclaration(attrName: InstanceIdentifier, attrType: DataType, isNullable: Boolean): Unit = {
-    out.puts(s"private ${kaitaiType2JavaTypeBoxed(attrType)} ${idToStr(attrName)};")
+  override def instanceDeclaration(instName: InstanceIdentifier, attrType: DataType, isNullable: Boolean): Unit = {
+    out.puts(s"private ${kaitaiType2JavaTypeBoxed(attrType)} ${idToStr(instName)};")
   }
 
   override def instanceWriteFlagDeclaration(attrName: InstanceIdentifier): Unit = {
--- a/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
@@ -67,6 +67,21 @@ class JavaCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     out.puts
   }
 
+  override def classAnnotation(spec: ClassSpec): Unit = {
+    if (!spec.isTopLevel) return;
+
+    importList.add("io.kaitai.struct.annotations.Generated")
+
+    out.puts("@Generated(")
+    out.inc
+    out.puts(s"""id = "${spec.name.last}",""")
+    out.puts(s"""version = "${KSVersion.current}",""")
+    out.puts(s"""posInfo = ${config.readStoresPos},""")
+    out.puts(s"""autoRead = ${config.autoRead}""")
+    out.dec
+    out.puts(")")
+  }
+
   override def classHeader(name: String): Unit = {
     val staticStr = if (out.indentLevel > 0) {
       "static "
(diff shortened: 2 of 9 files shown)


shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
```diff
@@ -419,7 +419,8 @@ class PythonCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
 
   override def switchEnd(): Unit = {}
 
-  override def instanceHeader(className: String, instName: InstanceIdentifier, dataType: DataType, isNullable: Boolean): Unit = {
+  override def instanceDeclaration(instName: InstanceIdentifier, instSpec: InstanceSpec): Unit = {}
+  override def instanceHeader(className: String, instName: InstanceIdentifier, instSpec: InstanceSpec): Unit = {
     out.puts("@property")
     out.puts(s"def ${publicMemberName(instName)}(self):")
```
KOLANICH: For python you need to
1. `import typing` to the top of a file where imports are situated
2. Output type of a function is shown by a `->` before `:`
3. Types of arguments are shown by adding `: TypeOfArg` right after arg name
4. Union types are `typing.Union[T1, T2]`

Mingun: Just for better understanding, can you post expected result?

KOLANICH: Damn, the comment I have written is irrelevant here.

Mingun:
I finish work on the feature. Now
- each generated class has `Generated` annotation,
- each field and getter, that represents attribute from `seq` element has `SeqItem` annotation,
- each field and getter, that represents instance from `instances` element has `Instance` annotation
- each field and getter, that represents parameter has `Parameter` annotation

PR with adds annotations to runtime -- https://github.com/kaitai-io/kaitai_struct_java_runtime/pull/23.

I wasn't ran tests with `kaitai_struct/tests/run-java`, because of lacking testng dependency (I don't known, why test script can not find it) but, because annotations do not change behavior of compiled class, I think, that nothing was break. I hope, CI do their work for me.

Mingun:
I would be happy to see it merged, as it will allow to proceed the improvement of `kaitai_struct_gui`. Runtime changes already merged, but their useless without support in compiler

Mingun:
@GreyCat, can you look at this? I didn't miss anything?

Mingun:
Can this be merged before another conflict occurred? There is no point in having support for annotation in the runtime without their support in the compiler

Mingun:
@GreyCat , please do not bury the project. Can you give feedback for my PRs?

GreyCat:
Hi! Apologies for super late response. This PR changes a lot, and I'm not 100% with you in certain aspects of it. I'll try to highlight main points of concern.

shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
```diff
@@ -91,14 +91,14 @@ class ClassCompiler(
...
-      List(
+    val allAttrs: List[(MemberSpec, Int)] =
+      curClass.seq.zipWithIndex ++
+      curClass.params.zipWithIndex ++
+      (List(
         AttrSpec(List(), RootIdentifier, CalcUserType(topClassName, None)),
         AttrSpec(List(), ParentIdentifier, curClass.parentType)
       ) ++
-      ExtraAttrs.forClassSpec(curClass, lang)
+      ExtraAttrs.forClassSpec(curClass, lang)).zipWithIndex
```
GreyCat: This looks like a huge artificial and redundant hack — i.e. even for attributes which clearly have no index, we still add some fictional index.

Mingun: This is only required for type compliance. As you said, many changes are required, but they are needed because I have to change the interface to supply the missing information. I tried to locate such changes in a separate commit, so it is best to review each commit separately.

In this case, these indexes are not really used. Since I do not know Scala, I decided not to introduce new interface methods, but to transmit all information through already existing methods.

So I can't forget to do something important if it's done in the process of calling these methods.

shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
```diff
@@ -184,28 +184,28 @@ class ClassCompiler(
...
-    attrs.foreach { (attr) =>
+  def compileAttrDeclarations(attrs: List[(MemberSpec, Int)]): Unit = {
+    attrs.foreach { case (attr, i) =>
       val isNullable = if (lang.switchBytesOnlyAsRaw) {
         attr.isNullableSwitchRaw
       } else {
         attr.isNullable
       }
-      lang.attributeDeclaration(attr.id, attr.dataTypeComposite, isNullable)
+      lang.attributeDeclaration(attr, i, isNullable)
```
GreyCat: Several concerns with this one are:

* It bundles two things in one `attributeDeclaration` call — actual attribute declaration + rendering of annotation. For docstrings, for example, we do them separately. Any chance you'd be ok to create something like `lang.attributeAnnotation(...)` where you would pass `i` and invoke it alongside this `attributeDeclaration` call from `ClassCompiler`?
* This annotation-generation behavior should be configurable (i.e. enable/disable switch). It makes sense to check it here and avoid call to `attributeAnnotation`-like procedure at all, rather than reimplementing that check in every single language.
* (Minor) This changes interface `attributeDeclaration` by adding `i` to it, which introduces a lot of unnecessary changes just due to refactoring. We can also avoid that by having distinct annotation rendering function (like `attributeAnnotation(...)`) rather than bundling it together.

shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
```diff
@@ -60,6 +60,10 @@ class JavaCompiler(val typeProvider: ClassTypeProvider, config: RuntimeConfig)
     // Used in every class
     importList.add(s"io.kaitai.struct.$kstructName")
     importList.add(s"io.kaitai.struct.$kstreamName")
+    importList.add("io.kaitai.struct.annotations.Generated")
+    importList.add("io.kaitai.struct.annotations.Instance")
+    importList.add("io.kaitai.struct.annotations.Parameter")
+    importList.add("io.kaitai.struct.annotations.SeqItem")
```
GreyCat: Again, this should be enabled by config flag when annotations are requested, not imported unconditionally.

shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
```diff
@@ -197,12 +210,55 @@ class JavaCompiler(val typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
+    annotate(attr, index)
+    out.puts(s"private ${kaitaiType2JavaType(attr.dataTypeComposite, isNullable)} ${idToStr(attr.id)};")
+  }
+
+  override def attributeReader(attr: MemberSpec, index: Int, isNullable: Boolean): Unit = {
+    annotate(attr, index)
+    out.puts(s"public ${kaitaiType2JavaType(attr.dataTypeComposite, isNullable)} ${idToStr(attr.id)}() { return ${idToStr(attr.id)}; }")
+  }
+
+  def doc(doc: DocSpec): String = {
```
GreyCat: Two concerns for this one:

* `doc` is a not a ideal naming idea to use in such a broad context. Please use some sort of descriptive name for that, such as `escapeDocAsString` or something like that.
* This escaping procedure actually exists already as [doStringLiteral](https://github.com/kaitai-io/kaitai_struct_compiler/blob/564aa4b5616a16a6c1598f622655b371b4393ddf/shared/src/main/scala/io/kaitai/struct/translators/CommonLiterals.scala#L11) or similar ASCII/Unicode-specific versions of it. It is used in very similar manner, for example, [for JSON serialization](https://github.com/kaitai-io/kaitai_struct_compiler/blob/564aa4b5616a16a6c1598f622655b371b4393ddf/shared/src/main/scala/io/kaitai/struct/JSON.scala#L30).

GreyCat: ... or, better yet, given that it's *Java*, it makes sense to call JavaTranslator's (i.e. `translator`) method.

shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
```diff
@@ -197,12 +210,55 @@ class JavaCompiler(val typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
+      }
+      sb.toString
+    }).getOrElse("")
   }
 
-  override def attributeReader(attrName: Identifier, attrType: DataType, isNullable: Boolean): Unit = {
-    out.puts(s"public ${kaitaiType2JavaType(attrType, isNullable)} ${idToStr(attrName)}() { return ${idToStr(attrName)}; }")
+  def annotate(attr: MemberSpec, index: Int) = {
+    attr match {
+      case param : ParamDefSpec => {
```
GreyCat: Everything else in the codebase uses `case foo: Foo` style, not `case foo : Foo`.

shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
```diff
@@ -197,12 +210,55 @@ class JavaCompiler(val typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
+      }
+      case field : AttrSpec => {
+        field.id match {
+          case NamedIdentifier(name) =>
+            out.puts(s"""@SeqItem(id = "${name}", index = ${index}, doc = "${doc(field.doc)}")""")
+          case NumberedIdentifier(_) =>
+            out.puts(s"""@SeqItem(id = "", index = ${index}, doc = "${doc(field.doc)}")""")
+          case _ => {}
+        }
+      }
```
GreyCat: You don't really need `{}` around block after `case ... => `

shared/src/main/scala/io/kaitai/struct/languages/PHPCompiler.scala
```diff
@@ -152,21 +152,15 @@ class PHPCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
-  override def attributeDeclaration(attrName: Identifier, attrType: DataType, isNullable: Boolean): Unit = {
-    attrName match {
-      case ParentIdentifier | RootIdentifier | IoIdentifier =>
-        // just ignore it for now
-      case _ =>
-        out.puts(s"protected $$_m_${idToStr(attrName)};")
-    }
-  }
+  override def attributeDeclaration(attr: MemberSpec, index: Int, isNullable: Boolean): Unit
+    = declaration(attr.id)
```
GreyCat: Nit: the rest of the codebase leaves `=` on the previous line, i.e.

```scala
def foo(foo: Foo): Foo =
  someFunction(foo)
```

shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
Mingun: > This annotation-generation behavior should be configurable

But in what case you may want to disable annotations generation? I consider them as part of the API interface, which implies that they are generated together with the attributes as their unenforceable component.

Although it is possible to distinguish the generation of annotations into a separate method, it seems to me that this will introduce an unnecessary assumption that in any language there may be annotations similar in syntax to annotations in Java, which may not be the case.

So, you still want that I split out annotations generation to new method?

GreyCat: Annotations carry a lot of information into .class file (like docstrings), which might not be what people want (for example, it will expose lots of internal ksy information into .class files, it will leak the docstrings, etc).

At the very least, C# [has them as "attributes"](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/), golang has [struct tags](https://golang.org/ref/spec#Struct_types) available via reflection (although they're much simpler and are string-only), [Rust has attributes](https://doc.rust-lang.org/reference/attributes.html), etc.

Even if that wasn't the case, the general idea is to keep `languages/` as simple and as template-like as possible. Ideally, it should output one simple thing, and we should not (ab)use stuff like "class header" or "attribute reader method header" to add tons of extra stuff there.

GreyCat:
I've took liberty to resolve the conflict which arisen due to recent changes. Please pull it first if you'll be contributing more stuff on top of this branch?

shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
Mingun: Ok, then I suggest to allow disable only doc-string generation. Position and original ID should not take much space and have to considered, as important part of implementation which allow to correctly visualize data. And `@Generated` annotation also contains some important stuff, like generator version. Then it will be even better to extract the documentation to a separate annotation:
```java
@Doc("Some documentation")// can be enabled by option, default is disabled
@SeqItem(pos = 1, id="some_field")// always generated
public int someField() { ... }
```

Mingun:
Actually, I plan to make a rebase to keep history linear as much as possible

Mingun:
I fully rework implementation:
* Remove generating `doc`
* Generate annotations via separate methods in `LanguageCompiler`, which by default do nothing
* Import annotation classes only when its used (annotations always generated though)
* Supply indexes (via `zipWithIndex`) only for that members for which its required

Because that includes also changes in runtime, I try to make PR for main repository: https://github.com/kaitai-io/kaitai_struct/pull/801. But as it my first experience with submodules PR, I don't know will it success or not. Therefore, if PR to the main repository fails to marge, tell me about it.

<details>
<summary>Generated file <code>params_enum.ksy</code></summary>

Source: https://github.com/kaitai-io/kaitai_struct_tests/blob/master/formats/params_enum.ksy
```java
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

package io.kaitai.struct.testformats;

import io.kaitai.struct.ByteBufferKaitaiStream;
import io.kaitai.struct.KaitaiStruct;
import io.kaitai.struct.KaitaiStream;
import java.io.IOException;
import io.kaitai.struct.annotations.Generated;
import java.util.Map;
import java.util.HashMap;
import io.kaitai.struct.annotations.Instance;
import io.kaitai.struct.annotations.Parameter;
import io.kaitai.struct.annotations.SeqItem;

@Generated(
    id = "params_enum",
    version = "0.9",
    posInfo = false,
    autoRead = true
)
public class ParamsEnum extends KaitaiStruct {
    public static ParamsEnum fromFile(String fileName) throws IOException {
        return new ParamsEnum(new ByteBufferKaitaiStream(fileName));
    }

    public enum Animal {
        DOG(4),
        CAT(7),
        CHICKEN(12);

        private final long id;
        Animal(long id) { this.id = id; }
        public long id() { return id; }
        private static final Map<Long, Animal> byId = new HashMap<Long, Animal>(3);
        static {
            for (Animal e : Animal.values())
                byId.put(e.id(), e);
        }
        public static Animal byId(long id) { return byId.get(id); }
    }

    public ParamsEnum(KaitaiStream _io) {
        this(_io, null, null);
    }

    public ParamsEnum(KaitaiStream _io, KaitaiStruct _parent) {
        this(_io, _parent, null);
    }

    public ParamsEnum(KaitaiStream _io, KaitaiStruct _parent, ParamsEnum _root) {
        super(_io);
        this._parent = _parent;
        this._root = _root == null ? this : _root;
        _read();
    }
    private void _read() {
        this.one = Animal.byId(this._io.readU1());
        this.invokeWithParam = new WithParam(this._io, this, _root, one());
    }
    public static class WithParam extends KaitaiStruct {

        public WithParam(KaitaiStream _io, Animal enumeratedOne) {
            this(_io, null, null, enumeratedOne);
        }

        public WithParam(KaitaiStream _io, ParamsEnum _parent, Animal enumeratedOne) {
            this(_io, _parent, null, enumeratedOne);
        }

        public WithParam(KaitaiStream _io, ParamsEnum _parent, ParamsEnum _root, Animal enumeratedOne) {
            super(_io);
            this._parent = _parent;
            this._root = _root;
            this.enumeratedOne = enumeratedOne;
            _read();
        }
        private void _read() {
        }
        @Instance(id = "is_cat")
        private Boolean isCat;
        @Instance(id = "is_cat")
        public Boolean isCat() {
            if (this.isCat != null)
                return this.isCat;
            boolean _tmp = (boolean) (enumeratedOne() == ParamsEnum.Animal.CAT);
            this.isCat = _tmp;
            return this.isCat;
        }
        @Parameter(index = 0, id = "enumerated_one")
        private Animal enumeratedOne;
        private ParamsEnum _root;
        private ParamsEnum _parent;
        @Parameter(index = 0, id = "enumerated_one")
        public Animal enumeratedOne() { return enumeratedOne; }
        public ParamsEnum _root() { return _root; }
        public ParamsEnum _parent() { return _parent; }
    }
    @SeqItem(index = 0, id = "one")
    private Animal one;
    @SeqItem(index = 1, id = "invoke_with_param")
    private WithParam invokeWithParam;
    private ParamsEnum _root;
    private KaitaiStruct _parent;
    @SeqItem(index = 0, id = "one")
    public Animal one() { return one; }
    @SeqItem(index = 1, id = "invoke_with_param")
    public WithParam invokeWithParam() { return invokeWithParam; }
    public ParamsEnum _root() { return _root; }
    public KaitaiStruct _parent() { return _parent; }
}
```
</details>

annotations

PR #192, opened 2020-01-15
Cast user type in switch with disabled autoRead

generalmimon:
Resolves https://github.com/kaitai-io/kaitai_struct/issues/204

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/CppCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/CppCompiler.scala
@@ -741,8 +741,14 @@ class CppCompiler(
     expr2
   }
 
-  override def userTypeDebugRead(id: String): Unit =
-    outSrc.puts(s"$id->_read();")
+  override def userTypeDebugRead(id: String, dataType: DataType, assignType: DataType): Unit = {
+    val expr = if (assignType != dataType) {
+      s"static_cast<${kaitaiType2NativeType(dataType)}>($id)"
+    } else {
+      id
+    }
+    outSrc.puts(s"$expr->_read();")
+  }
 
   override def switchRequiresIfs(onType: DataType): Boolean = onType match {
     case _: IntType | _: EnumType => false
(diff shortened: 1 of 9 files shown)


shared/src/main/scala/io/kaitai/struct/languages/CppCompiler.scala
```diff
@@ -741,8 +741,14 @@ class CppCompiler(
     expr2
   }
 
-  override def userTypeDebugRead(id: String): Unit =
-    outSrc.puts(s"$id->_read();")
+  override def userTypeDebugRead(id: String, dataType: DataType, assignType: DataType): Unit = {
+    val expr = if (assignType != dataType) {
+      s"static_cast<${kaitaiType2NativeType(dataType)}>($id)"
```
generalmimon: I frankly don't know how type castings in C++ work, can you check that line? I'd like to reuse the [`CppTranslator.doCast`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/a93c183af3c7d4137c1f7e6e0abb2813e149c883/shared/src/main/scala/io/kaitai/struct/translators/CppTranslator.scala#L157), which looks pretty solid, but I can't, because it expects `Ast.expr` and we currently don't have a reliable way how to convert language-specific code `string` to `Ast.expr` (this is actually the same problem that I faced in https://github.com/kaitai-io/kaitai_struct_compiler/pull/181, there I solved it by creating a special object `Ast.expr.CodeLiteral(String code)`, "translating" this AST object yields the `code` itself).

GreyCat: This should be fine, I guess, let's stick with this solution for now.

I don't really like the idea of `Ast.expr.CodeLiteral`, as this is effectively a hack which allows going target language string → AST → target language string for very specific scenarios, but we can definitely think more about what would be the best way to do it.

GreyCat:
Thanks, this looks like another massive effort!

PR #193, opened 2020-01-16
Don't prepend "src" to output dirs of Java and Go.

ams-tschoening:
Java and Go seem to be the only languages prepending "src" to their output directories, which make them fail in case a Maven-like directory tree with "src/main/java" is used. "java" itself needs to be the source folder already, not "java/src". Additionally, its easy for users to add paths on their own, but difficult to remove those added by KS.

Like has been discussed in the corresponding issue, "src" keeps getting added in case "--target all" is given, which is most likely only the case for tests and those use a dir structure of "java/bin|src", where that makes sense.

This fixes https://github.com/kaitai-io/kaitai_struct/issues/287.


GreyCat:
Thanks, look like a clean fix with a thought-out migration path!

PR #194, opened 2020-01-16
Adding GHI #287 to release notes, because it breaks builds.

ams-tschoening:
I've forgot about that in #193 and feel free to change wording of course. It just [shouldn't be missed](https://github.com/kaitai-io/kaitai_struct_compiler/pull/147#issuecomment-432764193) entirely.


GreyCat:
Thanks for your attention to details!

PR #195, opened 2020-01-22
Ignore and remove IntelliJ IDEA project files

dgelessus:
These files shouldn't be in the repo. Most of them are automatically generated by the IDEA Scala plugin when importing the sbt project, so tracking them in the repo is unnecessary and annoying (the plugin sometimes regenerates the IDEA files even if the sbt project hasn't changed, which leads to unnecessary changes/commits). The few non-auto-generated IDEA project files are either personal settings (e. g. project name) or factory defaults, which don't belong in the repo either.

The only downside of this change is that pulling it for the first time will break existing local IDEA projects. Not much can be done about this - though as a workaround you can make a backup copy of your .idea directory before pulling, and after pulling close the IDEA project, delete the broken .idea directory and replace it with the backup copy.


GreyCat:
I actually wonder why there's official JetBrains guidance to keep these files in the first place. What you mention here actually seems to make lots of sense. Thanks!

dgelessus:
> I actually wonder why there's official JetBrains guidance to keep these files in the first place.

If you have a project where the IntelliJ IDEs are the recommended (or required) development tools, it would make sense to share some common settings between developers (things like code style settings, inspection configuration, specific plugin settings, etc.). I'm guessing that this situation is more common in private/internal enterprise projects, because it's not something I see often in open-source projects. The only open-source project I know that *intentionally* contains IDEA project files is [IntelliJ itself](https://github.com/jetbrains/intellij-community).

Also, some projects don't use a proper build tool and instead rely on the IDE to compile and run the code, in which case the IDEA project is also your build configuration, so it needs to go into the repo. IMHO the proper fix for that is to use an actual build system (Maven, Gradle, sbt, whatever fits your JVM language).

In the case of ksc, none of this applies though - the build is done using sbt, and IDEA auto-generates its project configuration from the sbt build. In my experience so far, committing these auto-generated projects into version control does more harm than good - IDEA often doesn't recognize the existing generated project files, or it regenerates them with small changes for no apparent reason.

PR #196, opened 2020-02-21
Fix typo

cherue:
This changes the debug `ioOffset` from `KaitaiStream._byteOffset` (which is an implementation detail) to `KaitaiStream.byteOffset`. It's the same value (see [KaitaiStream.js#L85](https://github.com/kaitai-io/kaitai_struct_javascript_runtime/blob/7ea115fd84da0a459cde97d4cadd59a56d3df6c8/KaitaiStream.js#L85)) but with this change none of the underscore prefixed "internal" properties are used in the compiler. This makes writing other JS runtimes easier.


GreyCat:
Thanks!

PR #198, opened 2020-02-25
Fix processing in repetition

generalmimon:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/238 (+ tests [`process_repeat_bytes`](https://github.com/generalmimon/kaitai_struct_tests/blob/ec98f3eea9bb108ba9a92404763d3f87b42d1273/formats/process_repeat_bytes.ksy) and [`process_repeat_usertype`](https://github.com/generalmimon/kaitai_struct_tests/blob/ec98f3eea9bb108ba9a92404763d3f87b42d1273/formats/process_repeat_usertype.ksy)) for all languages (tried Go, Java, JavaScript, Python and PHP locally).

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/datatype/NeedRaw.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/NeedRaw.scala
@@ -1,0 +1,18 @@
+package io.kaitai.struct.datatype
+
+sealed trait NeedRaw {
+  val level: Int
+}
+
+case object NotRaw extends NeedRaw {
+  val level = 0
+}
+case object RawIo extends NeedRaw {
+  val level = 1
+}
+case object RawProcess extends NeedRaw {
+  val level = 1
+}
+case object RawIoProcess extends NeedRaw {
+  val level = 2
+}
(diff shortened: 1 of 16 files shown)


shared/src/main/scala/io/kaitai/struct/datatype/NeedRaw.scala
```diff
@@ -0,0 +1,18 @@
...
+}
+case object RawIo extends NeedRaw {
+  val level = 1
+}
+case object RawProcess extends NeedRaw {
+  val level = 1
+}
+case object RawIoProcess extends NeedRaw {
+  val level = 2
+}
```
GreyCat: Idea for improvement: define comparsion operators, so one can directly compare `RawIo < RawIoProcess`, etc.

shared/src/main/scala/io/kaitai/struct/languages/components/CommonReads.scala
```diff
@@ -83,12 +83,23 @@ trait CommonReads extends LanguageCompiler {
...
+      case utfb: UserTypeFromBytes => utfb.bytes.process.nonEmpty
+      case _ => false
+    }
+    (rawIo, rawProcess) match {
+      case (true, false) => RawIo
+      case (false, true) => RawProcess
+      case (true, true) => RawIoProcess
+      case _ => NotRaw
+    }
   }
```
GreyCat: Ideas for improvement: given that it only depends on `DataType`, make it `DataType` method — i.e. `dataType.needRaw` could return it.

PR #199, opened 2020-03-16
Simplify generation of fully-qualified class names in Python

dgelessus:
The generated names now look like `TopLevelType.MiddleType.NestedType` rather than `self._root.MiddleType.NestedType`.

It's not clear what the special case with `self._root` was meant to do. According to some old commit messages, it fixed some tests involving nested types, but these tests still pass without the special case.

This simplification has multiple advantages. Normal qualified class names (rather than strange `self._root` hacks) are more readable to humans and more understandable by IDEs and other tools. It also removes many unnecessary references to `self._root`, which allows parsing starting with a non-top-level type, e. g.:

```python
struct = TopLevelType.MiddleType.NestedType.from_io(stream)
```

With the old class name generation, this did not work, because the generated code expected that `_root` would always be an instance of the top-level type and not of a nested type.

(I discussed this issue on Gitter a few days ago: https://gitter.im/kaitai_struct/Lobby?at=5e64d8618011bb652afdc030)

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
@@ -534,12 +534,5 @@ object PythonCompiler extends LanguageCompilerStatic
     case _ => s"kaitaistruct.${err.name}"
   }
 
-  def types2class(name: List[String]): String = {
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks for investigation, let's give it a try!

PR #200, opened 2020-04-03
Regex support in valid key

jocelynke:
Working for python, cpp_stl_11. Have to be tested for other implementations (C#, Go, Java, JS, Ruby, Rust, PHP, Perl), not implemented on remaining. Can't work on cpp_stl_98 without other library.

### DIFF
--- a/Test.php
+++ b/Test.php
@@ -1,0 +1,20 @@
+<?php
+// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild
+
+namespace {
+    class Test extends \Kaitai\Struct\Struct {
+        public function __construct(\Kaitai\Struct\Stream $_io, \Kaitai\Struct\Struct $_parent = null, \Test $_root = null) {
+            parent::__construct($_io, $_parent, $_root);
+            $this->_read();
+        }
+
+        private function _read() {
+            $this->_m_name = \Kaitai\Struct\Stream::bytesToStr($this->_io->readBytes(2), "UTF-8");
+            if (!(preg_match("/\d\d/", $this->name()))) {
+                throw new \Kaitai\Struct\Error\ValidationRegexMatchError("\\d\\d", $this->name(), $this->_io(), "/seq/0");
+            }
+        }
+        protected $_m_name;
+        public function name() { return $this->_m_name; }
+    }
+}
--- a/shared/src/main/scala/io/kaitai/struct/datatype/KSError.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/KSError.scala
@@ -21,6 +21,7 @@ object KSError {
         case "ValidationLessThanError" => ValidationLessThanError
         case "ValidationGreaterThanError" => ValidationGreaterThanError
         case "ValidationNotAnyOfError" => ValidationNotAnyOfError
+        case "ValidationRegexMatchError" => ValidationRegexMatchError
       }
       excClass(dataType)
   }
@@ -62,6 +63,14 @@ case class ValidationGreaterThanError(_dt: DataType) extends ValidationError(_dt
   */
 case class ValidationNotAnyOfError(_dt: DataType) extends ValidationError(_dt) {
   def name = "ValidationNotAnyOfError"
+}
+
+/**
+  * Error to be thrown when validation fails with actual not matching regex
+  * @param _dt data type used in validation process
+  */
+case class ValidationRegexMatchError(_dt: DataType) extends ValidationError(_dt) {
+  def name = "ValidationRegexMatchError"
 }
 
 /**
--- a/shared/src/main/scala/io/kaitai/struct/format/ValidationSpec.scala
+++ b/shared/src/main/scala/io/kaitai/struct/format/ValidationSpec.scala
@@ -10,6 +10,7 @@ case class ValidationMax(max: Ast.expr) extends ValidationSpec
 case class ValidationRange(min: Ast.expr, max: Ast.expr) extends ValidationSpec
 case class ValidationAnyOf(values: List[Ast.expr]) extends ValidationSpec
 case class ValidationExpr(checkExpr: Ast.expr) extends ValidationSpec
+case class ValidationRegex(checkExpr: String) extends ValidationSpec
 
 object ValidationEq {
   val LEGAL_KEYS = Set("eq")
@@ -68,16 +69,37 @@ object ValidationExpr {
     }
 }
 
+object ValidationRegex {
+  val LEGAL_KEYS = Set("regex")
+
+  def fromMap(src: Map[String, Any], path: List[String]): Option[ValidationRegex] =
+    /*val regex = ParseUtils.getOptValueExpression(src, "regex", path)
+    regex match {
+      case Some(regexCt) => 
+        ParseUtils.ensureLegalKeys(src, LEGAL_KEYS, path)
+        Some((ValidationRegex(Ast.expr.Str(regexCt))))
+      case None => None
+    }
+  }*/
+    ParseUtils.getOptValueStr(src, "regex", path).map { case eqExpr =>
+      ParseUtils.ensureLegalKeys(src, LEGAL_KEYS, path)
+      ValidationRegex(eqExpr)
+    }
+}
+
+
 object ValidationSpec {
   val LEGAL_KEYS =
     ValidationEq.LEGAL_KEYS ++
     ValidationRange.LEGAL_KEYS ++
     ValidationAnyOf.LEGAL_KEYS ++
-    ValidationExpr.LEGAL_KEYS
+    ValidationExpr.LEGAL_KEYS ++
+    ValidationRegex.LEGAL_KEYS
 
   def fromYaml(src: Any, path: List[String]): ValidationSpec = {
     src match {
       case value: String =>
+
         fromString(value, path)
       case x: Boolean =>
         fromString(x.toString, path)
@@ -108,6 +130,9 @@ object ValidationSpec {
     val opt4 = ValidationExpr.fromMap(src, path)
     if (opt4.nonEmpty)
       return opt4.get
+    val opt5 = ValidationRegex.fromMap(src, path)
+    if (opt5.nonEmpty)
+      return opt5.get
 
     // No validation templates matched, check for any bogus keys
     ParseUtils.ensureLegalKeys(src, LEGAL_KEYS, path)
--- a/shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala
@@ -50,6 +50,7 @@ class RustCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     importList.add("std::default::Default")
     importList.add("kaitai_struct::KaitaiStream")
     importList.add("kaitai_struct::KaitaiStruct")
+    importList.add("regex::Regex")
 
     out.puts
   }
--- a/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
@@ -102,6 +102,14 @@ abstract class BaseTranslator(val provider: TypeProvider)
           case (ltype, rtype) =>
             throw new TypeMismatchError(s"can't compare $ltype and $rtype")
         }
+      case Ast.expr.RegexMatch(str: Ast.expr, regex: String) => {
+        detectType(str) match {
+          case (_: StrType) =>
+            doRegexMatchOp(translate(str), doRegex(regex))
+          case _ =>
+            throw new TypeMismatchError(s"regex match need strings")
+        }
+      }
       case Ast.expr.BinOp(left: Ast.expr, op: Ast.operator, right: Ast.expr) =>
         (detectType(left), detectType(right), op) match {
           case (_: NumericType, _: NumericType, _) =>
@@ -179,6 +187,7 @@ abstract class BaseTranslator(val provider: TypeProvider)
 
   def doEnumByLabel(enumTypeAbs: List[String], label: String): String
   def doEnumById(enumTypeAbs: List[String], id: String): String
+  def doRegex(reg: String): String = reg
 
   // Predefined methods of various types
   def strConcat(left: Ast.expr, right: Ast.expr): String = s"${translate(left)} + ${translate(right)}"
--- a/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
@@ -253,6 +253,11 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     }
   }
 
+  override def doRegexMatchOp(str: String, regex: String): String = {
+    importList.add("regexp")
+    s"matched, _ := regexp.MatchString(`${regex}`, ${trStringLiteral(str)}); matched"
+  }
+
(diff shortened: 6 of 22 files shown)


Test.php
```diff
@@ -0,0 +1,20 @@
...
+        private function _read() {
+            $this->_m_name = \Kaitai\Struct\Stream::bytesToStr($this->_io->readBytes(2), "UTF-8");
+            if (!(preg_match("/\d\d/", $this->name()))) {
+                throw new \Kaitai\Struct\Error\ValidationRegexMatchError("\\d\\d", $this->name(), $this->_io(), "/seq/0");
+            }
+        }
+        protected $_m_name;
+        public function name() { return $this->_m_name; }
+    }
+}
```
GreyCat, line 20: This should not be part of this commit.

shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala
```diff
@@ -50,6 +50,7 @@ class RustCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     importList.add("std::default::Default")
     importList.add("kaitai_struct::KaitaiStream")
     importList.add("kaitai_struct::KaitaiStruct")
+    importList.add("regex::Regex")
```
GreyCat, line 53: I seriously doubt that we should import that unconditionally.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -253,6 +253,11 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     }
   }
 
+  override def doRegexMatchOp(str: String, regex: String): String = {
+    importList.add("regexp")
+    s"matched, _ := regexp.MatchString(`${regex}`, ${trStringLiteral(str)}); matched"
```
GreyCat, line 258: Will this fail if it will be used twice in the same function?

shared/src/main/scala/io/kaitai/struct/translators/PHPTranslator.scala
```diff
@@ -57,6 +57,12 @@ class PHPTranslator(provider: TypeProvider, config: RuntimeConfig) extends BaseT
 
   override def doName(s: String) = s"${Utils.lowerCamelCase(s)}()"
 
+  override def doRegexMatchOp(str: String, regex: String): String = {
+    s"""preg_match(\"${regex}\", ${str})"""
+  }
+
+  override def doRegex(reg: String) : String = s"/" + reg + "/"
```
generalmimon, line 64: I guess just doing `preg_match` here on the **exact same regex** that was specified in `valid/regex` key isn't what you wanted. Please take in mind that this just checks if the string contains some **substring** matching the regex, it does not ensure that the **whole string** matches the regex.

I think of the `valid/regex` key as the `pattern` attribute of HTML `<input>` tag, as you can [read here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#Constraint_validation), "in the case of pattern attribute, we are matching against the entire value, not just any subset, as if  a ^(?: were implied at the start of the pattern and )$ at the end."

Some example:
```php
var_dump(preg_match("/web/", "PHP is the web scripting language of choice."));
// -> yields int(1) (which loosely converts to `true`), because it really does contain substring "web"
```

shared/src/main/scala/io/kaitai/struct/format/ValidationSpec.scala
```diff
@@ -68,16 +69,37 @@ object ValidationExpr {
...
+
+  def fromMap(src: Map[String, Any], path: List[String]): Option[ValidationRegex] =
+    /*val regex = ParseUtils.getOptValueExpression(src, "regex", path)
+    regex match {
+      case Some(regexCt) => 
+        ParseUtils.ensureLegalKeys(src, LEGAL_KEYS, path)
+        Some((ValidationRegex(Ast.expr.Str(regexCt))))
+      case None => None
+    }
+  }*/
```
GreyCat, line 83: If these lines are not used, they should be deleted.

shared/src/main/scala/io/kaitai/struct/translators/JavaScriptTranslator.scala
```diff
@@ -74,6 +74,12 @@ class JavaScriptTranslator(provider: TypeProvider) extends BaseTranslator(provid
   override def enumToInt(v: expr, et: EnumType): String =
     translate(v)
 
+  override def doRegexMatchOp(str: String, regex: String): String = {
+    s"RegExp(${regex}).test(${str})"
```
generalmimon, line 78: Please use the literal notation instead of using the `RegExp` constructor function. From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Creating_a_regular_expression):

> Regular expression literals provide compilation of the regular expression when the script is loaded. If the regular expression remains constant, using this can improve performance.

> Use the constructor function when you know the regular expression pattern will be changing, or you (...) are getting it from another source, such as user input."

I see no reason for using the constructor function because the `valid/regex` doesn't seem to support runtime-built regexes (they're always known in advance), so use the literal instead (just `/${regex}/`).

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
generalmimon: And I think it's not wise to ignore the error in case it occurs (doing `matched, _`). Please follow the same approach used in the rest of the `GoTranslator`, notice that the methods you added are ones of the few in the file returning `String` and not `TranslatorResult`. There is as special method `outVarCheckRes` handling the errors returned from functions, I suggest using that one.

shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala
jocelynke: Unlike other languages, the RustTranslator does not allow to "dynamically" add imports, how should I handle it ?

shared/src/main/scala/io/kaitai/struct/translators/PHPTranslator.scala
jocelynke: I had doubts about whether the user would have to specify ^ each time, or not, but I agree with your position, I'll check other languages as well.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
jocelynke: I tried to modify the translateExpr func in GoTranslator, but it seems to me that this function is not executed, am I wrong ?

generalmimon: > Will this fail if it will be used twice in the same function?

Yes, it probably will, because the variables can't be redeclared in the same scope in Go [unless some new variables are introduced](https://yourbasic.org/golang/redeclaring-variables/). I've made a simple test: https://repl.it/repls/WonderfulJointProcedure

> I tried to modify the translateExpr func in GoTranslator, but it seems to me that this function is not executed

@jocelynke Go language didn't play well with the classic `BaseTranslator` model that returns `String`s, because of the error handling in Go. Go doesn't have `try/catch` as other languages, it treats errors as values. This means that if you want the error from a function you call to bubble up in the call stack, it won't happen automatically. You have to check if the `error` value returned from the function is `nil` and if it's not, then an error occured and you should `return` it to the caller. (_Just returning it isn't exactly [debug-friendly](https://medium.com/rungo/error-handling-in-go-f0125de052f0#7fea), because the caller loses any context which function in the stack actually caused the error, but it's unfortunately done this way in KS currently._)

This means that parsing a single `u1` field from the stream can't be done in a single expression (because it can of course result in an EOF error), but instead it is a 3-phase process:

```go
func (this *HelloWorld) Read(...) (err error) {
 // ...
    tmp1, err := this._io.ReadU1()
    if err != nil {
        return err
    }
    this.One = tmp1
```

Because of this, `GoTranslator` can't just extend `BaseTranslator` as other languages do, because the methods in  `BaseTranslator` have return type `String` which doesn't fit for Go (and it can't be changed in the derived classes). Not all methods in `GoTranslator` can be implemented in a single expression which can be passed as `String`, for these the return value of the called error-returning method is saved to a temporary variable (named `tmp1`, `tmp2`, ... to avoid name conflicts), the error is handled and only the "serial number" of the temp variable is returned. This is done by the method [`outVarCheckRes`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala#L449-L454).

Hence most methods in `GoTranslator` return `TranslatorResult`, which is defined on the top:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/ef7178cb2e1dae8c443214e825bcf361e1e158ee/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala#L11-L13

And now how you should do it. I think the `doRegexMatchOp` here could look like this:

```scala
override def doRegexMatchOp(str: String, regex: String): TranslatorResult = {
  importList.add("regexp")
  outVarCheckRes(s"regexp.MatchString(`${regex}`, ${trStringLiteral(str)})")
```

Then you'll have to move the abstract definition of this method from `CommonOps` to `BaseTranslator`:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/7120c8ec65c7b5d35dc5c17333dace47470010ce/shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala#L68

The reason is that `CommonOps` is inherited by `GoTranslator` too (whereas `BaseTranslator` is not), but in `GoTranslator` you need to specify a different return type, which won't work in Scala. And the `Ast.expr.RegexMatch` handling in `BaseTranslator`

https://github.com/kaitai-io/kaitai_struct_compiler/blob/7120c8ec65c7b5d35dc5c17333dace47470010ce/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala#L105-L112

needs to be in `GoTranslator` as well, because again, it doesn't extend `BaseTranslator`.

jocelynke: Thank you for this deep explanation, I had implemented what you told me, but the thing is I have no log of a call to translateExpr in the GoTranslator. The consequence is that no code is generated to validate the field that needs it.

shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala
generalmimon: @jocelynke Ah, you mean that the `importList` is not passed as parameter to the translator as in other languages, then just adjust the constructor signature of `RustTranslator` and add it. Pretty similar to what I've done in https://github.com/kaitai-io/kaitai_struct_compiler/commit/0c02aec204787106d50040a203f52166d84ceca4 for C++.

But please bear in mind that the Rust spec generator of the KST translator (in the [`kaitai_struct_tests`](https://github.com/kaitai-io/kaitai_struct_tests) repo) has to be changed too, because it depends on the `RustTranslator` as well: [`RustSG.scala:13`](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/RustSG.scala#L13). This is almost always forgotten. Just open another PR for it and link to it from here.

generalmimon:
@jocelynke Could you please push your changes to your branch [`jocelynke:regex_support`](https://github.com/jocelynke/kaitai_struct_compiler/tree/regex_support) so we can see the changes you've made? It seems that you're just marking the things as resolved, but you don't publish anything so we can't say. The commits you push into this branch will appear in this PR.

jocelynke:
@generalmimon I wanted to correct everything before pushing. Here are the remaining to do : Go support, Rust dynamic include

shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala
jocelynke: Ok I've added it. Do you think it's better to open a new or continue the one I had already in the tests repo ?

generalmimon: > Do you think it's better to open a new or continue the one I had already in the tests repo ?

Please open a new one, https://github.com/kaitai-io/kaitai_struct_tests/pull/86 is not much related to KST translator updates.

jocelynke: Done : [kaitai-io/kaitai_struct_tests#87](https://github.com/kaitai-io/kaitai_struct_tests/pull/87)

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
jocelynke: @generalmimon I found that the valid key is not supported by Go in kaitai, so it's normal that no code is generated...

shared/src/main/scala/io/kaitai/struct/datatype/KSError.scala
```diff
@@ -64,6 +65,14 @@ case class ValidationNotAnyOfError(_dt: DataType) extends ValidationError(_dt) {
   def name = "ValidationNotAnyOfError"
 }
 
+/**
+  * Error to be thrown when validation fails with actual not matching regex
+  * @param _dt data type used in validation process
+  */
+case class ValidationRegexMatchError(_dt: DataType) extends ValidationError(_dt) {
```
GreyCat, line 72: For equality checks (and storing original value in question for C++), we need to know original value data type. But regular expression matches would likely only work for strings. What's the point in passing `_dt: DataType` here?

shared/src/main/scala/io/kaitai/struct/format/ValidationSpec.scala
```diff
@@ -68,16 +69,29 @@ object ValidationExpr {
...
+  val LEGAL_KEYS = Set("regex")
+
+  def fromMap(src: Map[String, Any], path: List[String]): Option[ValidationRegex] =
+    ParseUtils.getOptValueStr(src, "regex", path).map { case eqExpr =>
+      ParseUtils.ensureLegalKeys(src, LEGAL_KEYS, path)
+      ValidationRegex(eqExpr)
+    }
+}
+
+
```
GreyCat, line 82: Nit: unnecessary double newline.
```suggestion
```

shared/src/main/scala/io/kaitai/struct/format/ValidationSpec.scala
```diff
@@ -68,16 +69,29 @@ object ValidationExpr {
...
     ValidationRange.LEGAL_KEYS ++
     ValidationAnyOf.LEGAL_KEYS ++
-    ValidationExpr.LEGAL_KEYS
+    ValidationExpr.LEGAL_KEYS ++
+    ValidationRegex.LEGAL_KEYS
 
   def fromYaml(src: Any, path: List[String]): ValidationSpec = {
     src match {
       case value: String =>
+
```
GreyCat, line 94: Nit: unnecessary newline

```suggestion
```

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
```diff
@@ -198,4 +206,6 @@ abstract class BaseTranslator(val provider: TypeProvider)
   // for the language
   def anyField(value: Ast.expr, attrName: String): String =
     s"${translate(value)}.${doName(attrName)}"
+
+   def doRegexMatchOp(str: String, regex: String): String
```
GreyCat, line 210: Nit: formatting.

```suggestion
  def doRegexMatchOp(str: String, regex: String): String
```

PR #201, opened 2020-04-03
Fix invocation of nested types with parameters

fudgepop01:
Closes kaitai-io/kaitai_struct#724

done by changing the `ReUserTypeWithArgs` regex from `^([a-z][a-z0-9_]*)\((.*)\)$` to `^((?:[a-z][a-z0-9_]*(?:::)?)+)\((.*)\)$`

for a visual representation of what this regex does, see here:
https://regex101.com/r/3VY3DV/1

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
@@ -294,7 +294,7 @@ object DataType {
   private val ReIntType = """([us])(2|4|8)(le|be)?""".r
   private val ReFloatType = """f(4|8)(le|be)?""".r
   private val ReBitType = """b(\d+)""".r
-  private val ReUserTypeWithArgs = """^([a-z][a-z0-9_]*)\((.*)\)$""".r
+  private val ReUserTypeWithArgs = """^((?:[a-z][a-z0-9_]*(?:::)?)+)\((.*)\)$""".r
 
   def fromYaml(
     dto: Option[String],


generalmimon:
@fudgepop01 Thanks for the fix! It works fine and solves the problem.

generalmimon:
@fudgepop01 Well, I was a bit hasty in merging this PR without careful checking if it works right. First, it allows trailing double colon (e.g. `foo::bar::(5)`), which should be forbidden. And the second, much more serious issue, is that the regex is subject to [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html). It means that the repetition operators are nested in such a way that there are multiple ways to match the same match. In your particular regex, the problem is caused by the conditionality of the double colon `(?:::)`**`?`**, because it's contained in the group `(?:[a-z][a-z0-9_]*(?:::)?)+`, which is **repeated** using **`+`**. That means, according to the current "mood" of the regex engine, the double colon may be matched by this group (`(?:[a-z][a-z0-9_]*`**`::`**`)+`), but it also **doesn't have to be** (`(?:[a-z][a-z0-9_]*)+`). That's the cause of the problem. A string `abc` can be matched as `(abc)`, `(a)(bc)`, `(ab)(c)` or `(a)(b)(c)` (the number of options grows exponentially). And while the regex engine is backtracking and still cannot find the match, you can be sure that it tries all possible options. Try matching `aaaaaaaaaaaaa` against your regex at https://regex101.com/r/3VY3DV/2 (I don't recommend to do it anywhere else, you might not be able to terminate it the soft way) and then try adding more characters to it, you can see that every time you add a single letter, the number of steps approximatelly doubles, and the evaluation takes longer and longer. From 18 characters, it finally prints _catastrophic backtracking_, as the number of steps would exceed 1 000 000.

You can see the [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html) problem in action in the issue https://github.com/kaitai-io/kaitai_struct/issues/736#issue-605920646 that I created. The vulnerability caused by this is called [ReDoS](https://en.wikipedia.org/wiki/ReDoS) (regular expression denial of service) and it's quite common across many libraries. Fortunately, there are some handy tools that can detect the vulnerable regexes. I tried https://github.com/NicolaasWeideman/RegexStaticAnalysis, it seems to work pretty well.

I've already fixed this regex in https://github.com/kaitai-io/kaitai_struct_compiler/commit/6f6cc0c52c165a0d93c564f57dbf844de8a57b8f, but please be more careful next time. It was my fault as well that I didn't caught this, at least I learned something new about the regexes. It turns out that it may be quite a dangerous toy if you don't know what you're doing.

GreyCat:
Thanks for following up on that and fixing! I've missed this one as well, and I haven't really thought of a possiblity of DoS in this case.

Eventually we might want to migrate this to reuse KS expression parser elements instead of regexp, but so far so good ;)

PR #203, opened 2020-07-14
Correctly distinguish between floating point number and ineger following by attribute access

Mingun:
Fixes #175. Originally found when porting tests for [ksc-rs](https://github.com/Mingun/ksc-rs), literally that test:
https://github.com/kaitai-io/kaitai_struct_compiler/blob/05a6207d519bc8f14e0a2353a1bd5e0c3efb022d/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala#L380-L382

You can see that error in action with following KSY:
```yaml
meta:
  id: int2str
instances:
  value:
    value: 123.to_s
```


GreyCat:
This was a long outstanding bug that I've never found time or motivation to drill down into and fix :) Thanks so much for going this extra mile!

GreyCat:
Verified now, merging in! Thanks again for your effort!

PR #205, opened 2020-07-29
Fix array subscript ownership

generalmimon:
Resolves https://github.com/kaitai-io/kaitai_struct/issues/561 (merging this should fix the test `repeat_until_calc_array_type`, which is the last failing test related to that issue). 

I've discovered this C++11 bug when fixing test [`expr_calc_array_ops.ksy`](https://github.com/kaitai-io/kaitai_struct_tests/blob/9b170bbc2ee360a8980bd607751dd87c77c56eaf/formats/expr_calc_array_ops.ksy) for C++11 and during testing if changes in commit https://github.com/kaitai-io/kaitai_struct_compiler/commit/eb420a617ae91a803659dfc29c8a8b54516b7b70 work properly. The main problem encountered by test `expr_calc_array_ops` in C++11 was that KSC treated all resulting types of value instances as non-owning:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/22bc40e07ad7ad72e4f21fc165e21bbb54f35e0e/shared/src/main/scala/io/kaitai/struct/precompile/ValueTypesDeriver.scala#L23-L28

This fits most cases, but it doesn't work for array literals. If an array is created and stored in a value instance, the value instance must obviously own this array. This is already properly handled by the `detectTypeRaw` method: using `Ast.expr.List` would detect type `ArrayTypeInStream`, which is the owning type:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/eb420a617ae91a803659dfc29c8a8b54516b7b70/shared/src/main/scala/io/kaitai/struct/translators/TypeDetector.scala#L120-L124

So I've removed the rough conversion of all value instance types to non-owning and rather added the `.asNonOwning` call to the individual cases in [`detectTypeRaw`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/eb420a617ae91a803659dfc29c8a8b54516b7b70/shared/src/main/scala/io/kaitai/struct/translators/TypeDetector.scala#L37) method. The only AST objects that might return owning types (and must be converted to owning in any expression where they are used) currently are `Ast.expr.Name`, `Ast.expr.Subscript` and `Ast.expr.Attribute`. (`Ast.expr.List` is the only case where it is desirable.) So I'm converting the output types to non-owning from these operations.

However, the array subscripting operator makes it more complex. Subscripting an array of owning pointers (let's call it `my_array`) will yield the owning pointer itself (when the compiler does just `my_array()->at(0)` for C++11 as it normally does), and that causes a problem when you want to save it directly in some variable (imagine a value instance `value: my_array[0]`). It has to be converted to a non-owning pointer before copying it to any variable.

This .ksy sample I've created for testing shows the possible situations in a bit more detail:

```yaml
# Tests C++ ownership detection of arrays and their elements in value instances
meta:
  id: expr_array_usertype_ref
seq:
  - id: baz # unique_ptr<foo>
    type: foo
  - id: a # unique_ptr<vector<unique_ptr<foo>>>
          # Result: owning array with owning inner type
    type: foo
    repeat: expr
    repeat-expr: 1
types:
  foo:
    seq:
      - id: bar
        type: u1
instances:
  ut_array: # foo* (baz()) -> unique_ptr<vector<foo*>>>
            #
            #   Conclusion:
            #     - the value instance can have owning type (!!!)
            #     - the resulting type of any Ast.expr.Name is non-owning
            #   Result: owning array with non-owning inner type
    value: '[baz]'
  ut_array_ref: # vector<foo*>*
                # Result: non-owning array with non-owning inner type
    value: ut_array
  a_ref: # vector<unique_ptr<foo>>*
         #
         #   Conclusion: the resulting type of Ast.expr.Name must be non-owning, but the inner type of the array stays as is
         #   Result: non-owning array with owning inner type
    value: a
  a_sub: # vector<unique_ptr<foo>>* ->  unique_ptr<foo>  ->         foo*
         #          a()             ->     a()->at(0)    ->  a_ref()->at(0).get()
         #
         #   Conclusion: array subscript operation might return the owning type, but as such can't be saved directly to any
         #     variable (as we'd try to make a copy of unique_ptr) - we have to detect this situation and immediately convert
         #     it to non-owning pointer using .get()
    value: a[0]
  a_first: # Same as `a_sub`, just note that this is an alternative way how to subscript the array
    value: a.first
```

So the compiler needs to track whether the result of an expression originated as a subscript from an array with owning inner type. If so, then the `CppCompiler` has to use `.get()` on it before copying it to the `m_*` property. I call the property `isOwningInExpr`, because it sort of states if the type behaves as owning in expressions. I'm open to other suggestions.

I've noticed later that the test [`repeat_until_calc_array_type.cpp:64`](https://github.com/kaitai-io/ci_targets/blob/d3c61a1a71b415b4163ca7fee607d151410e7fb7/compiled/cpp_stl_11/repeat_until_calc_array_type.cpp#L64) is also encountering this bug, but it doesn't seem to me that the test focuses on revealing it. So this behavior should be rather covered by another test, the `expr_array_usertype_ref` as seen above can serve as a basis for it, but I'm not sure if there is any way how to improve it.

Is there any flaw in my logic and/or implementation? It would be nice if we wouldn't have to track in the compiler whether to add `.get()` or not, but I don't know C++11 well enough to come up with another solution.


generalmimon:
Actually, I've just realized that all what can be done using value instances can be also done by directly constructing arguments of parameter types (so parameters can own an array when created directly in the argument: `type: param_type([foo, bar]), it might be needed to convert the type having `isOwningInExpr == true` to non-owning pointer using `.get()`, etc.), so we need another test on that. Probably another changes will be needed to make it work, but it should be far easier when we already have the infrastructure for `isOwningInExpr` variable established for the value instances.

GreyCat:
Thanks very much @generalmimon, sorry for taking so much time to get to review this one. I took liberty at fixing the merge conflict and I confirm that it's indeed a working solution.

I concur that it looks like not a very elegant one, but for now, I believe we should definitely merge it.

PR #206, opened 2020-08-10
lua5.1 compatibility

smarek:


shared/src/main/scala/io/kaitai/struct/translators/LuaTranslator.scala
```diff
@@ -22,15 +24,31 @@ class LuaTranslator(provider: TypeProvider, importList: ImportList) extends Base
     '\33' -> "\\027"
   )
 
-  override def strLiteralUnicode(code: Char): String =
-    "\\u{%04x}".format(code.toInt)
+  override def strLiteralUnicode(code: Char): String = {
+    // compatible unicode encoding UTF8 bytes as \ddd where d = number 0-255
+    // handles both 1 and 2 bytes Chars/codes
```
generalmimon, line 29: A single UTF-8 code point can occupy up to 4 bytes, see page [UTF-8 on Wikipedia](https://en.wikipedia.org/wiki/UTF-8#Description). So writing `1 and 2 bytes` is at least misleading.

I'm also not sure what you've meant with `Chars/codes`. Why is `Chars` written with the first uppercase letter, is it a name?

Did you mean _characters_ and _code points_? If so, I suppose that the term _code point_ is more suitable here, see [this page](https://www.twilio.com/docs/glossary/what-utf-8#utf-8-basics):

> Instead of _characters_, it is actually more correct to refer to [_code points_](https://en.wikipedia.org/wiki/Code_point) when discussing encoding systems. Code points allow abstraction from the term _character_ and are the atomic unit of storage of information in an encoding. Most code points represent a single character, but some represent information such as formatting.

generalmimon (changes requested):

shared/src/main/scala/io/kaitai/struct/translators/LuaTranslator.scala
```diff
@@ -22,15 +24,31 @@ class LuaTranslator(provider: TypeProvider, importList: ImportList) extends Base
...
 
   override def numericBinOp(left: Ast.expr, op: Ast.operator, right: Ast.expr) = {
     (detectType(left), detectType(right), op) match {
       case (_: IntType, _: IntType, Ast.operator.Div) =>
         s"math.floor(${translate(left)} / ${translate(right)})"
       case _ =>
-        super.numericBinOp(left, op, right)
+        op match {
+          case Ast.operator.BitAnd =>
+            s"bit.band(${translate(left)}, ${translate(right)})"
```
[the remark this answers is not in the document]
smarek: Done

shared/src/main/scala/io/kaitai/struct/translators/LuaTranslator.scala
smarek: Yup, i have these messed up, but obviously removing my code comments would be the best, as the code works

smarek:
Well I'm happy you appreciated my work, i've updated the commit to reflect on your points.

Also it would be great, if the PRs on this repository built with Travis, so we know the PR is working well (eg. the mentioned conflict about used Java APIs when building ScalaJS build for JavaScript).

PR #208, opened 2020-08-18
Check Python runtime version using tuple comparison instead of string parsing

dgelessus:
Part of one of the potential fixes for kaitai-io/kaitai_struct#804. Depends on kaitai-io/kaitai_struct_python_runtime#49.


generalmimon (approved):
@dgelessus Looks great, thanks! 💯

generalmimon:
Well, I swear I knew a few minutes ago that I need to bump the versions by one (0.8->0.9 and 0.9->0.10) before merging this 😄, but no big deal, I'll do that right away.

PR #210, opened 2020-11-18
Replace octal constants with unicode constants.

Mingun:
Using octal constants is the hard error at least in the Scala 2.13.3


Mingun:
Is there build checks on Circle CI is even run? It seems no...

generalmimon (approved):
LGTM, thanks for the contribution!

PR #211, opened 2020-11-18
Fix build on Java 13-14

Mingun:
Recent comments in the https://github.com/sbt/sbt/issues/5093 say that problem is solved from Scala 2.12.10, so update to the last 2.12 release


Mingun:
This is the next step after #215 -- it's fixes pipeline

generalmimon (approved):
Thanks!

PR #212, opened 2020-11-19
Fix generation of the negative long literals

Mingun:
This change fixes at least the https://github.com/kaitai-io/kaitai_struct_tests/blob/master/formats/enum_long_range_s.ksy test

Also, @GreyCat, what the trick that you have done in the dbda9b765e1c57f8e2e5dcfcf7d33dc7b1e8e48e?
https://github.com/kaitai-io/kaitai_struct_compiler/blob/f19304de3ce0ef6504dcb241e7847aa303684381/shared/src/main/scala/io/kaitai/struct/translators/JavaTranslator.scala#L13-L17

I would appreciate if you can drop a comment explaining this to the codebase

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/translators/JavaTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/JavaTranslator.scala
@@ -10,12 +10,23 @@ import io.kaitai.struct.languages.JavaCompiler
 
 class JavaTranslator(provider: TypeProvider, importList: ImportList) extends BaseTranslator(provider) {
   override def doIntLiteral(n: BigInt): String = {
+    // Java's integer parsing behaves differently depending on whether you use decimal or hex syntax.
+    // With decimal syntax, the parser/compiler rejects any number that cannot be stored in a long
+    // (signed 64-bit integer) without overflow. With hexadecimal syntax, it allows anything that
+    // would fit in an unsigned 64-bit integer. Java's `long` is always signed, so if you use this
+    // trick to enter a number that is too large for a signed 64-bit integer, it will overflow into
+    // negative. But this can still be useful if you don't perform any arithmetic that cares about
+    // the sign (e. g. you pass the value unmodified to something else, or you use only bit operations
+    // or Long's "unsigned" methods).
+    //
+    // Of course, if `n > Long.MaxValue * 2 + 1` we'll still get out of range error
+    // TODO: Convert real big numbers to BigInteger
     val literal = if (n > Long.MaxValue) {
       "0x" + n.toString(16)
     } else {
       n.toString
     }
-    val suffix = if (n > Int.MaxValue) "L" else ""
+    val suffix = if (n < Int.MinValue || n > Int.MaxValue) "L" else ""
 
     s"$literal$suffix"
   }


shared/src/main/scala/io/kaitai/struct/translators/JavaTranslator.scala
```diff
@@ -15,7 +15,7 @@ class JavaTranslator(provider: TypeProvider, importList: ImportList) extends Bas
     } else {
       n.toString
     }
-    val suffix = if (n > Int.MaxValue) "L" else ""
+    val suffix = if (Int.MinValue > n || n > Int.MaxValue) "L" else ""
```
generalmimon, line 18: This looks better:

```suggestion
    val suffix = if (n < Int.MinValue || n > Int.MaxValue) "L" else ""
```

generalmimon (approved):

shared/src/main/scala/io/kaitai/struct/translators/JavaTranslator.scala
Mingun: For taste and color...

generalmimon:
> Also, @GreyCat, what the trick that you have done in the [dbda9b7](https://github.com/kaitai-io/kaitai_struct_compiler/commit/dbda9b765e1c57f8e2e5dcfcf7d33dc7b1e8e48e)?

I think this is purely a matter of code purity. Big integers written in decimal notation like `9223372036854775807` may be hard to read for us humans, and in binary formats is often the hex notation more convenient. Just guessing :)

In any case, `n > Long.MaxValue` is an error, such large `n` cannot be written just using the long `(\d+)L` literal without the use of [`BigInteger`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/math/BigInteger.html). Perhaps it's supposed to be `n > Int.MaxValue` so that it makes sense :)

Mingun:
Yes, but maybe this is some workaround for BigInt->Long conversion in Scala

generalmimon:
> Yes, but maybe this is some workaround for BigInt stringification in Scala

I don't think so, there's naturally no difference in calling `.toString` with no argument (which presumes base 10) and `.toString(16)`:

```scala
// Start writing your ScalaFiddle code here
def dumpBigInt(n: BigInt): Unit = {
  println("(n > Long.MaxValue): " + (n > Long.MaxValue))
  println("n: " + n.toString)
  println("n: 0x" + n.toString(16))
}

dumpBigInt(BigInt(    "7fff"+"ffff"+"ffff"+"ffff", 16))
dumpBigInt(BigInt("1"+"0000"+"0000"+"0000"+"0000", 16))
```

Here's the output of the above code [in ScalaFiddle](https://scalafiddle.io/sf/RLnpAx1/0):

```
(n > Long.MaxValue): false
n: 9223372036854775807
n: 0x7fffffffffffffff
(n > Long.MaxValue): true
n: 18446744073709551616
n: 0x10000000000000000
```

It seems to be just a matter of preference, like I said in https://github.com/kaitai-io/kaitai_struct_compiler/pull/212#issuecomment-730399454. There's no point in looking for a deeper reason I think :)

dgelessus:
Java's integer parsing behaves differently depending on whether you use decimal or hex syntax. With decimal syntax, the parser/compiler rejects any number that cannot be stored in a `long` (signed 64-bit integer) without overflow. With hexadecimal syntax, it allows anything that would fit in an *unsigned* 64-bit integer. Java's `long` is always signed, so if you use this trick to enter a number that is too large for a *signed* 64-bit integer, it will overflow into negative. But this can still be useful if you don't perform any arithmetic that cares about the sign (e. g. you pass the value unmodified to something else, or you use only bit operations or `Long`'s "unsigned" methods).

```java
jshell> 18446744073709551615L
|  Error:
|  integer number too large
|  18446744073709551615L
|  ^

jshell> 0xffffffffffffffffL
$1 ==> -1

jshell> import java.math.BigInteger

jshell> new BigInteger("18446744073709551615", 10).equals(new BigInteger("ffffffffffffffff", 16))
$2 ==> true
```

generalmimon:
@dgelessus 
> Java's integer parsing behaves differently depending on whether you use decimal or hex syntax.

I see, thanks for the explanation. Let's write it into the comment in the code!

Mingun:
Added a comment

generalmimon:
> Java's `long` is always signed, so if you use this trick to enter a number that is too large for a _signed_ 64-bit integer, it will overflow into negative. But this can still be useful if you don't perform any arithmetic that cares about the sign (e. g. you pass the value unmodified to something else, or you use only bit operations or **`Long`'s "unsigned" methods**).

We could create a test for these unsigned calculations, at least as a reminder that this problem exists in Java.

It seems that there's a lot of fun with Java integers:

1. the "always signed" concept requires using the higher sizes for unsigned types (`s1` -> `byte`, `u1` -> `short`, `u2` -> `int`, `u4` -> `long`),
2. boxed integers can't be compared without type casting or perhaps the [`compareTo`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Long.html#compareTo(java.lang.Long)) method ([`ExprIfIntOps.java:47-56`](https://github.com/kaitai-io/ci_targets/blob/28ec4c14ab93c0f97d726e5f6ccf0426d132a213/compiled/java/src/io/kaitai/struct/testformats/ExprIfIntOps.java#L47-L56)),
3. now I know that hex and decimal long literals are different,
4. there is an alternative set of unsigned operations on boxed integers like [`compareUnsigned​`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Long.html#compareUnsigned(long,long)), [`parseUnsignedLong​`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Long.html#parseUnsignedLong(java.lang.String)), [`toUnsignedString`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Long.html#toUnsignedString(long)), ...

What else can we expect in the future 🥱?

dgelessus:
> 2\. boxed integers can't be compared without type casting to primitive types

You can compare boxed and unboxed numbers using primitive operators - automatic boxing and unboxing has been supported since Java 6 (I think):

```java
jshell> Long.valueOf(123) == 123L
$1 ==> true
```

Although you need to be careful when comparing two boxed numbers, because then `==` and `!=` *don't* auto-unbox and instead compare the boxed numbers by object identity. This can give misleading results, because it's possible to have two boxed numbers that are not the same object but have the same value:

```java
jshell> Long.valueOf(12345678) == Long.valueOf(12345678)
$2 ==> false
```

For added confusion, the non-equality operators (`<`, `<=`, `>=`, `>`) do operate on the number's value - because these operators are not defined for reference types, auto-unboxing *does* happen here:

```java
jshell> Long.valueOf(12345678) <= Long.valueOf(12345678)
$3 ==> true

jshell> Long.valueOf(12345678) < Long.valueOf(12345678)
$4 ==> false
```

> 3\. boxed types cannot be casted directly to a primitive type with different byte size

Also not quite true - integers (boxed or unboxed) are cast automatically to the needed size, but only if it's safe (i. e. the target type is at least as large as the source type):

```java
jshell> (long)Short.valueOf((short)123)
$5 ==> 123

jshell> (short)Long.valueOf(456)
|  Error:
|  incompatible types: java.lang.Long cannot be converted to short
|  (short)Long.valueOf(456)
|         ^---------------^
```

The unfortunate side effect of this is that because all Java integer literals are `int` by default, you sometimes have to add manual casts in contexts where a `short` is required specifically, like in the example above.

IMO it's a much bigger issue that boxed types can be nullable, because that's the one case where the auto-unboxing simplification *doesn't* work - when a `null` boxed number is passed into code that does auto-unboxing, you suddenly gett `NullPointerException`s from what looks like a simple and safe primitive calculation.

> 5\. there is an alternative set of unsigned operations on boxed integers

Small correction: these operations actually operate on unboxed primitive numbers - they are just declared as static methods on the boxed type, because Java doesn't support methods on primitive types.

> It seems that there's a lot of fun with Java integers

You are entirely correct with that though. 🙂

Mingun:
Is there any something that need to do? Otherwise its ready to merge

Mingun:
@GreyCat, @generalmimon, @dgelessus yet another suspended discussion. Can we finish it in this year?

Mingun:
Well... tomorrow is the anniversary of the last comment. Any hope or all is dust and air?

shared/src/main/scala/io/kaitai/struct/translators/JavaTranslator.scala
```diff
@@ -10,12 +10,18 @@ import io.kaitai.struct.languages.JavaCompiler
 
 class JavaTranslator(provider: TypeProvider, importList: ImportList) extends BaseTranslator(provider) {
   override def doIntLiteral(n: BigInt): String = {
+    // When we deal with large positive numbers we can't just write it as is,
+    // because `long` is signed and java compiler would complain us about this.
+    // But in many cases we can treat `long` as unsigned and work with it
+    // accordingly. In that cases we can write it in the hexadecimal form.
```
generalmimon, line 16: This comment doesn't look to me like it explains https://github.com/kaitai-io/kaitai_struct_compiler/pull/212#issuecomment-730541577 at all, it is written somewhat vaguely and clumsily.

I would be more happy to pretty much copy-paste @dgelessus's post on this (https://github.com/kaitai-io/kaitai_struct_compiler/pull/212#issuecomment-730541577) or perhaps just link to it from the code, that post explains it _way_ better than this code comment in my opinion.

generalmimon (changes requested):

dgelessus (approved):
LGTM - can confirm that this indeed fixes the enum_long_range_s test. Sorry for the long wait!

PR #215, opened 2020-11-19
Use GitHub Actions for run tests

Mingun:


.github/workflows/scala.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [1.8, 11, 14, 15]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK ${{matrix.java}}
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{matrix.java}}
```
generalmimon, line 19: I prefer to put spaces around the variables in GH Actions config, but that's just my personal preference:

```suggestion
    - name: Set up JDK ${{ matrix.java }}
      uses: actions/setup-java@v1
      with:
        java-version: ${{ matrix.java }}
```

.github/workflows/scala.yml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Run tests
+      run: sbt test
```
generalmimon, line 21: I'm afraid that `sbt` does not come from the `actions/setup-java` action - searching for [`sbt` in its repo](https://github.com/actions/setup-java/search?q=sbt&type=code) found nothing.

generalmimon (changes requested):
Thanks for this contribution, it'd be nice to have it, but there are some issues:

.github/workflows/scala.yml
Mingun: This is just default template that GitHub suggested me and I only was add several versions of java. I think, that template was manually written instead of auto-generated so it should work. I didn't get a chance to check if it is work, though

Mingun: I make a test and `sbt` already exists in the container:
https://github.com/Mingun/kaitai_struct_compiler/runs/1428552667

.github/workflows/scala.yml
```diff
@@ -0,0 +1,21 @@
+name: Scala CI
```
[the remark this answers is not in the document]
Mingun: Done

.github/workflows/scala.yml
Mingun: Done

Mingun:
Is there any something that need to do? Otherwise its ready to merge

generalmimon (approved):
LGTM, thanks for the contribution!

PR #218, opened 2020-11-22
Publish unit test results

Mingun:
This is attempt to show results of test runs in the PRs. I don't know how it will look, because can't test it due to nature of the `pull_request_target` event: it will use workflow file from the base repository (i.e. unmodified from `master` branch). And without `pull_request_target` event PRs `GITHUB_TOKEN`s have no permissions to comment on PRs. So feel free to close this PR if it is unwanted behaviour.

Fixes https://github.com/kaitai-io/kaitai_struct/issues/830

PR #219, opened 2020-11-22
Show test results in console

Mingun:
Another attempt to fix https://github.com/kaitai-io/kaitai_struct/issues/830 (the other one is #218).

Just output test result to the console, but you will also see it when you run `sbt test` locally. I do not think that this is a big problem, on the contrary, it seems like a plus, but I am not a Scala developer and do not know how it will be convenient for Scala developers, and also I did not find how to override the output settings from the command line.

So feel free to close this PR if it is unwanted behavior.


generalmimon (approved):
Thanks!

PR #222, opened 2020-11-24
Fix scala.js deprecation warnings

Mingun:


js/src/main/scala/io/kaitai/struct/MainJs.scala
```diff
@@ -9,7 +9,7 @@ import scala.scalajs.js
 import scala.scalajs.js.JSConverters._
 import scala.scalajs.js.annotation.JSExport
 
-@JSExport
+@JSExportTopLevel
 object MainJs {
```
Mingun, line 13: I don't known is that enough. Type of the global `MainJs` in JS changes from factory function to object itself. However, I couldn't find the code that would use `MainJs`

js/src/main/scala/io/kaitai/struct/MainJs.scala
```diff
@@ -7,9 +7,9 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.scalajs.js
 import scala.scalajs.js.JSConverters._
-import scala.scalajs.js.annotation.JSExport
+import scala.scalajs.js.annotation.{JSExport, JSExportTopLevel}
 
-@JSExport
+@JSExportTopLevel("MainJs")
 object MainJs {
```
Mingun, line 13: I don't known is that enough. Type of the global `MainJs` in JS changes from factory function to object itself. However, I couldn't find the code that would use `MainJs`

generalmimon: > However, I couldn't find the code that would use `MainJs`

What about this usage (I recommend using simple fulltext search instead of the "smart" Find Usages... features):

https://github.com/kaitai-io/kaitai_struct_compiler/blob/af69228e70cfdab96eb2fc334b9b0e8aab3c52aa/build.sbt#L210-L215

generalmimon:
I need to check this once I'm on my computer, if it works as before or not.

js/src/main/scala/io/kaitai/struct/MainJs.scala
Mingun: I've use simple fulltext search, because I'm not have any scala extensions installed in my VSCode. But I didn't notice that actually returned value is used several lines above as `factory` variable. I've fix that

Mingun:
Hey! Let's finish this (and [other PRs](https://github.com/pulls?q=is%3Aopen+is%3Apr+author%3AMingun+user%3Akaitai-io+archived%3Afalse)) before end of year! 😃 @GreyCat, @generalmimon, @dgelessus

generalmimon (changes requested):
If you compare your and upstream version of the generated bundle `js/npm/kaitai-struct-compiler.js` after running `sbt compile fastOptJS buildNpmJsFile`, you find that the path to the object `MainJs` has also changed:

```diff
diff --git "1/.\\kaitai-struct-compiler-upstream.js" "2/.\\kaitai-struct-compiler-Mingun-scala-js.js"
index 02dcd47..0f698e1 100644
--- "1/.\\kaitai-struct-compiler-upstream.js"
+++ "2/.\\kaitai-struct-compiler-Mingun-scala-js.js"
@@ -1,10 +1,10 @@
-(function (root, factory) {
+(function (root, compiler) {
   if (typeof define === 'function' && define.amd) {
-    define([], factory);
+    define([], function() { return compiler; });
   } else if (typeof module === 'object' && module.exports) {
-    module.exports = factory();
+    module.exports = compiler;
   } else {
-    root.KaitaiStructCompiler = factory();
+    root.KaitaiStructCompiler = compiler;
   }
 }(this, function () {
 
@@ -93251,10 +93251,7 @@ var $d_scm_ArrayBuffer = new $TypeData().initClass({
   Ljava_io_Serializable: 1
 });
 $c_scm_ArrayBuffer.prototype.$classData = $d_scm_ArrayBuffer;
-$e.io = ($e.io || {});
-$e.io.kaitai = ($e.io.kaitai || {});
-$e.io.kaitai.struct = ($e.io.kaitai.struct || {});
-$e.io.kaitai.struct.MainJs = $m_Lio_kaitai_struct_MainJs$;
+$e.MainJs = $m_Lio_kaitai_struct_MainJs$();
 }).call(this);
 //# sourceMappingURL=kaitai-struct-compiler-js-fastopt.js.map

return exports.io.kaitai.struct.MainJs;

}));
```

So `return exports.io.kaitai.struct.MainJs` here:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/d10b795ade2b5d0151c39c5a36cf94dea555495d/build.sbt#L213-L217

also needs to be changed into:

```js
        |$compiledFileContents 
        | 
        |return exports.MainJs; 
        | 
        |}));
```

Mingun:
I've found that probably changing `factory` to `compiler` is wrong. According my analysis, current `master` does the following:
- `root.KaitaiStructCompiler` is a _kaitai struct compiler constructor function_, should be used as `KaitaiStructCompiler()` or `new KaitaiStructCompiler()` which gives us a _compiler object_
- `module.exports` the same as `root.KaitaiStructCompiler`
- `define` accepts a factory that produces _kaitai struct compiler constructor function_

After upgrading scala-js and do not touching `factory` stuff:
- `root.KaitaiStructCompiler` is a _compiler object_ itself, should be used as `KaitaiStructCompiler`
- `module.exports` the same as `root.KaitaiStructCompiler`
- `define` accepts a factory that produces a _compiler object_

Such a change seems reasonable and I think we shouldn't does anything with that. As I see, nobody in kaitai_struct repos using this generated file, but I've not checking web-IDE

generalmimon:
> Such a change seems reasonable and I think we shouldn't does anything with that.

Well, I wholeheartedly agree, as the stuff like

```js
const KaitaiStructCompiler = require('kaitai-struct-compiler');
const compiler = new KaitaiStructCompiler();
// compiler.compile(...);
```

that shows the common usage at the moment actually doesn't make any sense - `KaitaiStructCompiler` is just some 0-arg trivial function like `function() { return MainJs; }` so invoking it with `new` is just pure nonsense, and the whole step `compiler = KaitaiStructCompiler()` is completely redundant as well.

But I'm reluctant to make such a daring BC break, because [`kaitai-struct-compiler` at _npm_](https://www.npmjs.com/package/kaitai-struct-compiler) is used quite a lot in the wild (as you can see from the download stats, it has ~300 downloads per week, whatever that means) and this change is guaranteed to break **all** code using it.

> As I see, nobody in kaitai_struct repos using this generated file, but I've not checking web-IDE

Well, at the _very least_, GitHub so-called _dependents_ by scanning the `package.json` files: https://github.com/kaitai-io/kaitai_struct_compiler/network/dependents

And yeah, the Web IDE is essentially based on the Scala.js bundle of KSC, I think that actually the Web IDE was the reason why the Scala.js target of the KSC was introduced in the first place :)

So we've got a dilemma here of simplifying the usage but breaking the BC and just leaving it as it is for now (in the not-so-ideal state). I mean - the fix is incredibly simple (you just remove the calling brackets `()`), but you still have to know that you need to do it, and it can be really annoying to be an unsuspecting user, run `npm update` and wonder why the app just stopped working. But yeah, I know - we need to do the BC breaks sometime 🙂

I see that you added an info to the README, that's great, but I wonder whether we can also automatically show a warning in the console right after a user updates the `kaitai-struct-compiler` package to the version where the BC break happens. Would something like this work?

`warn-about-factory-bc-break.js`
```js
console.warn(
'kaitai-struct-compiler: WARNING - you\'ve just installed the 0.10.0-SNAPSHOT20201217+ version. \
If you\'re coming from an older version, make sure to update your code - \
the usage changed from "(new KaitaiStructCompiler()).compile(...)" to just "KaitaiStructCompiler.compile(...)".\n'
);
```

`package.json`
```json
{
  "version": "0.10.0-SNAPSHOT20201217-...",
  "scripts": {
    "postinstall": "node warn-about-factory-bc-break.js"
  }
}
```

I think that this warning can be quite useful for the users.

RELEASE_NOTES.md
```diff
@@ -1,3 +1,7 @@
+# 0.11 (TBA)
```
generalmimon, line 1:
```suggestion
# 0.10 (TBD)
```

js/README.md
```diff
@@ -53,23 +53,24 @@ Our [examples repository](https://github.com/kaitai-io/kaitai_struct_examples) c
 
 ## Plugging in
 
+> :information_source: Before the version 0.11.0 the `kaitai-struct-compiler` module returned a _compiler constructor_ (called `KaitaiStructCompiler`) and you used it as `new KaitaiStructCompiler()`, but since 0.11.0 release this is the _compiler object_ itself and using `new` is not required anymore
```
generalmimon, line 56:
```suggestion
> :information_source: Before the version 0.10.0 the `kaitai-struct-compiler` module returned a _compiler constructor_ (called `KaitaiStructCompiler`) and you used it as `new KaitaiStructCompiler()`, but since 0.10.0 release this is the _compiler object_ itself and using `new` is not required anymore
```

RELEASE_NOTES.md
```diff
@@ -1,3 +1,7 @@
...
+* Packaging / infrastructure improvements:
+  * [npm package](https://www.npmjs.com/package/kaitai-struct-compiler/) now returns the compiler object itself instead of constructor function ([#222](https://github.com/kaitai-io/kaitai_struct_compiler/pull/222))
```
generalmimon, line 3:
```suggestion
  * [npm package `kaitai-struct-compiler`](https://www.npmjs.com/package/kaitai-struct-compiler/) now returns the compiler object itself instead of constructor function ([#222](https://github.com/kaitai-io/kaitai_struct_compiler/pull/222))
```

Mingun:
> But I'm reluctant to make such a daring BC break, because [`kaitai-struct-compiler` at _npm_](https://www.npmjs.com/package/kaitai-struct-compiler) is used quite a lot in the wild (as you can see from the download stats, it has ~300 downloads per week, whatever that means) and this change is guaranteed to break **all** code using it.

Well, `kaitai-struct-compiler` follows semver semantics, so that's not a big problem. Until version 1.0 breaking changes is allowed when changing minor version. In the long-term perspective is is better to fix such kind of things before we reaching version 1.0 because versions 0.x are just for this! Anyway, in production code you shouldn't depends on the unstable tags, so if you depends on something like `latest` version, I think, you are understands all pros & cons.


> So we've got a dilemma here of simplifying the usage but breaking the BC and just leaving it as it is for now (in the not-so-ideal state). I mean - the fix is incredibly simple (you just remove the calling brackets `()`), but you still have to know that you need to do it, and it can be really annoying to be an unsuspecting user, run `npm update` and wonder why the app just stopped working. But yeah, I know - we need to do the BC breaks at some point 🙂

I definitely sure that we should do that. In any case when you upgrade your dependencies to next minor/major version you shouldn't expect that all will work as before 🙂



> I see that you added an info to the README, that's great, but I wonder whether we can also automatically show a warning in the console right after a user updates the `kaitai-struct-compiler` package to the version where the BC break happens. Would something like this work?

I've add this to the PR, but I think it will not be so useful, because it seems that this script will not run on `npm update` but these users is the main contingent that should be notified. I was unable to find instructions to display a message when `npm update`, not when `npm install`. It is possible, of course, that the update includes installation... For that reason I've add this warning as separate commit, so If after all you decide that it useless, just skip it when merge.

js/README.md
```diff
@@ -53,23 +53,24 @@ Our [examples repository](https://github.com/kaitai-io/kaitai_struct_examples) c
 
 ## Plugging in
 
+> :information_source: Before the version 0.10.0 the `kaitai-struct-compiler` module returned a _compiler constructor_ (called `KaitaiStructCompiler`) and you used it as `new KaitaiStructCompiler()`, but since 0.10.0 release this is the _compiler object_ itself and using `new` is not required anymore (see [#222](https://github.com/kaitai-io/kaitai_struct_compiler/pull/222)).
```
generalmimon, line 56:
```suggestion
> :information_source: Before the version 0.10.0 the `kaitai-struct-compiler` module returned a _compiler constructor_ (called `KaitaiStructCompiler`) and you would use it as `(new KaitaiStructCompiler()).compile(...)`, but since 0.10.0 release it returns the _compiler object_ itself and you just need to do `KaitaiStructCompiler.compile(...)`. Make sure to adapt your code.
```

js/warn-about-constructor-bc-break.js
```diff
@@ -0,0 +1,6 @@
+console.warn(
+'kaitai-struct-compiler: WARNING - you\'ve just installed the 0.10.0-SNAPSHOT20201217+ version. \
+If you\'re coming from an older version, make sure to update your code - \
+the usage changed from "(new KaitaiStructCompiler()).compile(...)" to "KaitaiStructCompiler.compile(...)".\n\n\
+See https://github.com/kaitai-io/kaitai_struct_compiler/pull/222\n'
```
generalmimon, line 5: IMHO a reference to this PR is not needed. There isn't any additional essential info here - the message already contains everything what a user needs to know.

generalmimon (changes requested):

js/warn-about-constructor-bc-break.js
```diff
@@ -0,0 +1,6 @@
...
+);
```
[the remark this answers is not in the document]
Mingun: Added

js/warn-about-constructor-bc-break.js
Mingun: Removed

Mingun:
Ping

generalmimon (approved):
@Mingun Thanks!

PR #227, opened 2020-12-01
Fix non-exhaustive warnings with the Ast.expr.Call node

Mingun:
Because function pointers are forbidden in the expression language from now parser
won't allow anymore following constructions:
- `42()`
- `"string"()`
- `true()`
- `(...)()`
- `[]()`


Mingun:
@GreyCat, @generalmimon, I see that you have some activity in the project recently. Could you find a time to review my PRs, for example, this?

PR #231, opened 2021-04-11
GoCompiler: Fix bug for member name is builtin keyword.

hluwa:
The executable/dex.ksy compiled code like it:
```go
type Dex struct {
    ...
    _f_map bool
    map *Dex_MapList
    ...
}
...
func (this *Dex) Map() (v *Dex_MapList, err error) {
    if (this._f_map) {
        return this.map, nil
    }
    ...
    this.map = tmp5
    ...
    return this.map, nil
}
```

The `map` is Golang built type, cannot name a member. After fix:
```go
type Dex struct {
    ...
    _f_map_ bool
    map_ *Dex_MapList
    ...
 }
....
func (this *Dex) Map() (v *Dex_MapList, err error) {
    if (this._f_map_) {
        return this.map_, nil
    }
    ...
    this.map_ = tmp5
    ...
    return this.map_, nil
}
```

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
@@ -522,13 +522,18 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     out.dec
     out.puts(")")
   }
+  val builtin: List[String] = List("var", "const", "type", "package", "import", "func",
+    "return", "defer", "go", "select", "interface", "struct",
+    "break", "case", "continue", "for", "fallthrough", "else", "if", "switch", "goto", "default",
+    "chan", "type", "map", "range", "panic")
 
+  def excludeKeyword(name: String): String = if (builtin.contains(name)) s"${name}_" else name
   def idToStr(id: Identifier): String = {
     id match {
       case SpecialIdentifier(name) => name
       case NamedIdentifier(name) => Utils.upperCamelCase(name)
       case NumberedIdentifier(idx) => s"_${NumberedIdentifier.TEMPLATE}$idx"
-      case InstanceIdentifier(name) => Utils.lowerCamelCase(name)
+      case InstanceIdentifier(name) => excludeKeyword(Utils.lowerCamelCase(name))
       case RawIdentifier(innerId) => "_raw_" + idToStr(innerId)
       case IoStorageIdentifier(innerId) => "_io_" + idToStr(innerId)
     }
@@ -550,7 +555,7 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
 
   override def localTemporaryName(id: Identifier): String = s"_t_${idToStr(id)}"
 
-  override def paramName(id: Identifier): String = Utils.lowerCamelCase(id.humanReadable)
+  override def paramName(id: Identifier): String = excludeKeyword(Utils.lowerCamelCase(id.humanReadable))
 
   def calculatedFlagForName(id: Identifier) = s"_f_${idToStr(id)}"
 


shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
```diff
@@ -522,13 +522,18 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
+  val builtin: List[String] = List("var", "const", "type", "package", "import", "func",
+    "return", "defer", "go", "select", "interface", "struct",
+    "break", "case", "continue", "for", "fallthrough", "else", "if", "switch", "goto", "default",
+    "chan", "type", "map", "range", "panic")
 
+  def excludeKeyword(name: String): String = if (builtin.contains(name)) s"${name}_" else name
   def idToStr(id: Identifier): String = {
     id match {
       case SpecialIdentifier(name) => name
       case NamedIdentifier(name) => Utils.upperCamelCase(name)
```
generalmimon, line 534: Not sure if this one should also be sanitized. I'll put a test KSY file together as @GreyCat suggested in https://github.com/kaitai-io/kaitai_struct/issues/90#issuecomment-311636287 (it will essentially look like [this one for Python](https://gist.github.com/Natronium/edcba05f7f474de150da472c1d1344b1) but also with added `seq` attributes I guess) so that we know for sure what all needs to be handled.

generalmimon (changes requested):

generalmimon:
```javascript
const reservedWords = ["var", "const", "type", "package", "import", "func",
    "return", "defer", "go", "select", "interface", "struct",
    "break", "case", "continue", "for", "fallthrough", "else", "if", "switch", "goto", "default",
    "chan", "type", "map", "range", "panic"];

let ksy = '';
ksy += (
`meta:
  id: golang_keyword_test
`
);

ksy += (
`instances:
` + reservedWords.map(
    (v, i) =>
`  ${v}:
    value: golang_keywords::${v}
`
  ).join('')
);

ksy += (
`enums:
  golang_keywords:
` + reservedWords.map(
    (v, i) =>
`    ${i}: ${v}
`).join('')
);

ksy += (
`types:
  seq_type:
    seq:
` + reservedWords.map(
    (v, i) =>
`      - id: ${v}
        size: 0
`
  ).join('')
);
console.log(ksy);
```

generalmimon:
The KSY generated by the above code already showed the problem with your list of reserved words (as defined at [`GoCompiler.scala:525-528`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/b3514debea54008a4c056f641a09dd35070e3929/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala#L525-L528)):

```
io.kaitai.struct.format.YAMLParseException: /types/seq_type/seq/23: duplicate attribute ID 'type', previously defined at /types/seq_type/seq/2
```

https://github.com/kaitai-io/kaitai_struct_compiler/blob/b3514debea54008a4c056f641a09dd35070e3929/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala#L525-L528

```diff
   val builtin: List[String] = List("var", "const", "type", "package", "import", "func",
     "return", "defer", "go", "select", "interface", "struct",
     "break", "case", "continue", "for", "fallthrough", "else", "if", "switch", "goto", "default",
-    "chan", "type", "map", "range", "panic")
+    "chan", "map", "range", "panic")
```

This also highlights the importance of the alphabetic order - we would immediately see that there are duplicate values because they would be right next to each other.

shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
generalmimon: Oh, I see - the convention for `seq` identifiers is `UpperCamelCase` for Go:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/b3514debea54008a4c056f641a09dd35070e3929/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala#L534

which don't suffer from this problem since Go is case sensitive and only lowercase words are reserved.

generalmimon:
Also, I noticed that there is one extra word that isn't present at https://golang.org/ref/spec#Keywords:

```javascript
// https://golang.org/ref/spec#Keywords
let goSpecReservedWords = new Set(
`break        default      func         interface    select
case         defer        go           map          struct
chan         else         goto         package      switch
const        fallthrough  if           range        type
continue     for          import       return       var`.match(/[a-z0-9]+/g)
);

// https://github.com/kaitai-io/kaitai_struct_compiler/blob/b3514de/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala#L525-L528
let hluwaReservedWords = new Set(["var", "const", "type", "package", "import", "func",
    "return", "defer", "go", "select", "interface", "struct",
    "break", "case", "continue", "for", "fallthrough", "else", "if", "switch", "goto", "default",
    "chan", "type", "map", "range", "panic"]);

// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#implementing_basic_set_operations
function difference(setA, setB) {
    let _difference = new Set(setA)
    for (let elem of setB) {
        _difference.delete(elem)
    }
    return _difference
}

console.log(difference(goSpecReservedWords, hluwaReservedWords));
// => Set(0) {}

console.log(difference(hluwaReservedWords, goSpecReservedWords));
// => Set(1) {"panic"}
```

According to [this page](https://notes.shichao.io/golang/#filenames-keywords-and-identifiers), `panic` is not a reserved word, but a _predeclared identifier_. It probably does not matter in terms of how we'll handle such names (although it might in theory - maybe predeclared identifiers _can_ be used normally in some special context?), but we'll want to at least make a clear distinction between them while defining the blacklist (i.e. don't merge them into one alphabetically sorted list). Either declare two lists or label two groups within one:

```scala
  // From https://notes.shichao.io/golang/#filenames-keywords-and-identifiers, sorted alphabetically
  val reservedWords: List[String] = List(
    // Reserved words
    "break", "case", "chan", "const", "continue",
    "default", "defer", "else", "fallthrough", "for",
    "func", "go", "goto", "if", "import",
    "interface", "map", "package", "range", "return",
    "select", "struct", "switch", "type", "var",

    // Predeclared identifiers
    "append", "bool", "byte", "cap", "close", "complex",
    "complex128", "complex64", "copy", "false", "float32", "float64",
    "imag", "int", "int16", "int32", "int64", "int8",
    "iota", "len", "make", "new", "nil", "panic",
    "print", "println", "real", "recover", "string", "true",
    "uint", "uint16", "uint32", "uint64", "uint8", "uintptr",
  )
```

hluwa:
Oh, Thank you for your work. I think is no need to deal with "Predefined Identifiers" , Because it can be named to the variable normally. I have updated the commit.
```scala
  // From https://golang.org/ref/spec#Keywords, sorted alphabetically
  val reservedWords: List[String] = List(
    // Reserved words
    "break", "case", "chan", "const", "continue",
    "default", "defer", "else", "fallthrough", "for",
    "func", "go", "goto", "if", "import",
    "interface", "map", "package", "range", "return",
    "select", "struct", "switch", "type", "var",
  )
```

PR #235, opened 2021-11-28
Fix constants generated in case of `parent: false` in Lua, Python and Ruby

Mingun:
- lua: false -> nil
- python: False -> None
- ruby: false -> nil

Perl probably also affected by this but I don't known what to use instead of `0`


dgelessus (approved):
Makes sense, thanks!

For Perl, my guess is that `undef` would be the right choice. That seems to be the closest thing to "null" that Perl has, and I *think* it's also what `parent` would be set to if you create a struct object without a parent. I don't know Perl very well though, so I'm not confident enough to make that change myself.

The fixes for the other affected languages are good, so I'll go ahead and merge this as-is. I'll leave the Perl fix to someone who knows the language better than I do :)

PR #238, opened 2021-12-11
Allow `repeat-expr` and `repeat-until` keys without a `repeat` key

Mingun:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/776


KOLANICH:
Why not

```yaml
repeat:
  expr: 10
```

instead?

Mingun:
The discussion about possible changes was so long ago that I completely forgot about this solution. Anyway, implementing completely new solution requires a bit more time and this change is simple and backward-compatible

PR #242, opened 2022-01-30
Generate imports for Construct

Mimickal:
This causes imports to be generated for the Construct target. This **does not** fix https://github.com/kaitai-io/kaitai_struct/issues/703, but it does make Construct behave the same as Python (for imports, anyway).

Given this ksy
```yaml
meta:
  id: example3
  endian: le
  imports:
    - common/other
seq:
  - id: field1
    type: other
```

We get this Construct (Python) output:
```python
from construct import *
from construct.lib import *

from other import *

example3 = Struct(
	'field1' / LazyBound(lambda: other),
)

_schema = example3
```


wbarnha:
Any update on this?

PR #244, opened 2022-02-21
add utf16 string types to golang generator

lucebac:
I needed UTF16 strings for some Kaitai structs and since Golang supports them similarly to the other encoding, I just added them to the generator.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
@@ -277,9 +277,8 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     "iso8859-4" -> ("charmap.ISO8859_4", IMPORT_CHARMAP),
     "sjis" -> ("japanese.ShiftJIS", "golang.org/x/text/encoding/japanese"),
     "big5" -> ("traditionalchinese.Big5", "golang.org/x/text/encoding/traditionalchinese"),
-    "utf16-auto" -> ("unicode.UTF16(unicode.LittleEndian, unicode.UseBOM)", "golang.org/x/text/encoding/unicode"),
-    "utf16-le" -> ("unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode"),
-    "utf16-be" -> ("unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode")
+    "utf-16le" -> ("unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode"),
+    "utf-16be" -> ("unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode")
   )
 
   override def bytesToStr(value: Ast.expr, expr: Ast.expr): TranslatorResult =
--- a/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
@@ -276,7 +276,10 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     "iso8859-3" -> ("charmap.ISO8859_3", IMPORT_CHARMAP),
     "iso8859-4" -> ("charmap.ISO8859_4", IMPORT_CHARMAP),
     "sjis" -> ("japanese.ShiftJIS", "golang.org/x/text/encoding/japanese"),
-    "big5" -> ("traditionalchinese.Big5", "golang.org/x/text/encoding/traditionalchinese")
+    "big5" -> ("traditionalchinese.Big5", "golang.org/x/text/encoding/traditionalchinese"),
+    "utf16-auto" -> ("unicode.UTF16(unicode.LittleEndian, unicode.UseBOM)", "golang.org/x/text/encoding/unicode"),
+    "utf16-le" -> ("unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode"),
+    "utf16-be" -> ("unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode")
   )
 
   override def bytesToStr(value: Ast.expr, expr: Ast.expr): TranslatorResult =


shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -276,7 +276,10 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     "iso8859-3" -> ("charmap.ISO8859_3", IMPORT_CHARMAP),
     "iso8859-4" -> ("charmap.ISO8859_4", IMPORT_CHARMAP),
     "sjis" -> ("japanese.ShiftJIS", "golang.org/x/text/encoding/japanese"),
-    "big5" -> ("traditionalchinese.Big5", "golang.org/x/text/encoding/traditionalchinese")
+    "big5" -> ("traditionalchinese.Big5", "golang.org/x/text/encoding/traditionalchinese"),
+    "utf16-auto" -> ("unicode.UTF16(unicode.LittleEndian, unicode.UseBOM)", "golang.org/x/text/encoding/unicode"),
+    "utf16-le" -> ("unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode"),
+    "utf16-be" -> ("unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)", "golang.org/x/text/encoding/unicode")
```
generalmimon, line 282: 1. "`utf16`" is almost certainly **not** the canonical name in most languages - in @GreyCat's table of encoding identifiers, it's referred to as "<code>UTF<strong>-</strong>16{,BE,LE}</code>" (notice the hyphen), the same for [IANA list of character sets](https://www.iana.org/assignments/character-sets/character-sets.xhtml), [RFC 2781](https://www.rfc-editor.org/rfc/rfc2781.html), list of supported encodings [in JavaScript](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings) (`utf16` is not even recognized as an encoding[^1]), [in Python](https://docs.python.org/3/library/codecs.html) (`utf16` is recognized, but ~~`utf16-be`~~ or ~~`utf16-le`~~ are not) etc.

   You must only use `utf-16`, `utf-16be` or `utf-16le`.

2. "`utf16-auto`" is controversial. The `-auto` suffix is almost certainly non-standard, I cannot find a single environment using such identifier. It should probably be simply "`utf-16`". But I'm not even sure about the behavior - you used `unicode.LittleEndian` as the default encoding, which will be used if BOM is not present (as governed by `unicode.UseBOM`) - see https://pkg.go.dev/golang.org/x/text@v0.3.7/encoding/unicode#UTF16:

   > For UseBOM, if there is no starting BOM, it will proceed with the default Endianness.

   **Should UTF-16LE be the default?** I checked [RFC 2781](https://datatracker.ietf.org/doc/html/rfc2781#section-4.3) and it suggests otherwise (yeah, it's an old memo, but how do you find up-to-date info of this kind):

   > If the first two octets of the text is not 0xFE followed by
   0xFF, and is not 0xFF followed by 0xFE, then the text SHOULD be
   interpreted as being big-endian.

   This looks funny, because not everyone seems to care about such recommendations - JavaScript's "`utf-16`" is an alias for "`utf-16le`". I don't know why that is, but hey (https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings):

   > Label | Encoding
   > -- | --
   > "`utf-16`", "`utf-16le`" | ['utf-16le'](https://en.wikipedia.org/wiki/UTF-16#Byte_order_encoding_schemes)

   So JavaScript behaves exactly like it MUST NOT behave 😞

   > Applications that process text with the "UTF-16"
   charset label MUST NOT assume the serialization without first
   checking the first two octets to see if they are a big-endian BOM, a
   little-endian BOM, or not a BOM.

   I haven't yet checked what other languages do when they decode UTF-16 from bytes without BOM, but it would be probably worth checking so that we know what to do.
   

[^1]: This JavaScript snippet
      ```js
      (new TextDecoder('utf16')).decode(new Uint8Array([0xFF, 0xFE, 0x3D, 0xD8, 0xDC, 0xDC]))
      ```
      raises the following error (the same goes for ~~`utf16-be`~~ and ~~`utf16-le`~~):
      ```
      RangeError: Failed to construct 'TextDecoder': The encoding label provided ('utf16') is invalid.
      ```"utf-16", "utf-16le"	['utf-16le'](https://en.wikipedia.org/wiki/UTF-16#Byte_order_encoding_schemes)

generalmimon (changes requested):

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
lucebac: Thanks for your comment - and I am sorry for the long delay. There were other projects which had a higher priority but finally I have the time to work on this again.

I removed the `-auto` encoding and adjusted the other names to what is expected by the standard. Since you edited your comment: is there anything left to do then? As far as I can see there was only the naming left, which should be fixed now. The rest was only caused by my erroneous naming.

generalmimon:
@lucebac Note: don't worry about failing tests, that's normal 🙂

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
generalmimon: > 2\. I haven't yet checked what other languages do when they decode UTF-16 from bytes without BOM, but it would be probably worth checking so that we know what to do. I've also found this official [FAQ about UTF-16 and BOM](https://www.unicode.org/faq/utf_bom.html) - however, this particular case is not included there.

I checked a few languages and the behavior of `UTF-16` encoding (without explicit byte order) is largely inconsistent. Most languages respect BOM if present ([Java](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html), [Python](https://docs.python.org/3/library/codecs.html#encodings-and-unicode)) but some do not ([JavaScript](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings), where `utf-16` is an alias for `utf-16le` - little-endian BOM is ignored and big-endian BOM is simply copied as if it were a normal character, but the bytes are still interpreted in LE order), [Python](https://docs.python.org/3/library/codecs.html#encodings-and-unicode) will use native endianness for `UTF-16` if there's no BOM, [Java](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html) defaults to big-endian if there is no BOM, etc.

<details>
<summary>Java code</summary>

```java
import java.io.UnsupportedEncodingException;

public class Main {
	public static void main(String[] args) throws UnsupportedEncodingException {
		// UTF-16BE without BOM
		System.out.println(new String(new byte[]{/*                  */0x30,0x53,0x30,(byte)0x93,0x30,0x6B,0x30,0x61,0x30,0x6F}, "UTF-16"));
		// UTF-16BE with BOM
		System.out.println(new String(new byte[]{(byte)0xFE,(byte)0xFF,0x30,0x53,0x30,(byte)0x93,0x30,0x6B,0x30,0x61,0x30,0x6F}, "UTF-16"));
		// UTF-16LE with BOM
		System.out.println(new String(new byte[]{(byte)0xFF,(byte)0xFE,0x53,0x30,(byte)0x93,0x30,0x6B,0x30,0x61,0x30,0x6F,0x30}, "UTF-16"));

		// Output: "こんにちは\n" three times
	}
}
```
</details>

<details>
<summary>Python code (Windows x86_64 => little-endian)</summary>

```python
# UTF-16BE with BOM
print(b'\xfe\xff0S0\x930k0a0o'.decode('utf-16'))
# UTF-16LE with BOM
print(b'\xff\xfeS0\x930k0a0o0'.decode('utf-16'))
# UTF-16LE without BOM
print(b'S0\x930k0a0o0'.decode('utf-16'))

# Output: "こんにちは\n" three times
```
</details>


<details>
<summary>JavaScript code</summary>

```javascript
var strUtf16Le = (new TextDecoder('utf-16')).decode(new Uint8Array([0xFF, 0xFE, 0x3D, 0xD8, 0xDC, 0xDC]))
// => '📜'
var strUtf16Be = (new TextDecoder('utf-16')).decode(new Uint8Array([0xFE, 0xFF, 0xD8, 0x3D, 0xDC, 0xDC]))
console.log(strUtf16Be)
// => '￾㷘�'
for (let codePoint of strUtf16Be) {
   console.log(codePoint.codePointAt(0).toString(16))
}
// => fffe, 3dd8, fffd
```
</details>

It would be nice to standardize it, but currently there's even not any format in [KSF repo](https://github.com/kaitai-io/kaitai_struct_formats) using `encoding: UTF-16` without explicit `LE` or `BE` order (except [`nt_mdt_pal.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/commit/8baf5315cc874510740ea61abbd6196687dac48e#commitcomment-69744833), where `encoding: UTF-16` is an error and it should be `UTF-16LE`), so there's probably no point in making such an effort. But we can at least warn about it or ban it altogether, as @GreyCat suggests in https://github.com/kaitai-io/kaitai_struct/issues/391#issuecomment-376003487.

generalmimon (approved):
Thanks for the contribution!

PR #248, opened 2022-06-21
Golang: fix enum type for large values

lucebac:
When using 32bit Golang, the standard `int` type only holds 32 bits. This is a problem when the enum keys are larger than `Int.MaxValue`. This fix should solve the problem for Golang. However, I cannot tell whether this issue also affects other languages which then need a fix as well. Maybe, there is a more thorough solution for this anyway - I'm not at all used to Scala so my fix might be not the ideal way of doing this.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/GoCompiler.scala
@@ -513,8 +513,13 @@ class GoCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     val fullEnumName: List[String] = curClass ++ List(enumName)
     val fullEnumNameStr = types2class(fullEnumName)
 
+    var enumType = "int"
+    if (enumColl.last._1 > Int.MaxValue) {
+        enumType = "int64"
+    }
+
     out.puts
-    out.puts(s"type $fullEnumNameStr int")
+    out.puts(s"type $fullEnumNameStr $enumType")
     out.puts("const (")
     out.inc
 


puhitaku:
+1

GreyCat:
@lucebac @puhitaku
Apologies for very late response.

1. Do we need a new test to cover these among others already available in [tests repo](https://github.com/kaitai-io/kaitai_struct_tests/tree/master/formats)?
2. Do we need a special golang build to test this? (rather than the [one we use in CI](https://github.com/kaitai-io/kaitai_struct_docker_images/tree/master/src/go/1.19-linux-x86_64))

lucebac:
1. I am not sure. Please see this part of my code where I needed these large enum values for: https://github.com/GDATAAdvancedAnalytics/winreg-tasks/blob/72bc92e0623f9ed3212a04bf81bcce9efc7d7216/kaitai/optional_settings.ksy#L89..L93
2. no, Golang natively supports 64 bit integers. My change just makes kaitai use 64bit integers in enums when the values exceed the maximum for 32bit integers. So as long as the fix is in place, everything should be working ootb

GreyCat:
> 1. I am not sure. Please see this part of my code where I needed these large enum values for: https://github.com/GDATAAdvancedAnalytics/winreg-tasks/blob/72bc92e0623f9ed3212a04bf81bcce9efc7d7216/kaitai/optional_settings.ksy#L89..L93

Got it, let me try to reproduce it in a test.

In this specific case you seem to be abusing enums for bitmasks. Very likely you'd want to replace that enum with a separate type with [bit-sized flags](https://doc.kaitai.io/user_guide.html#_bit_sized_integers).

lucebac:
Yes and no. I want to be able to export the enum members in generated code so I believe the enum is correct

lucebac:
Hi @GreyCat 
have found the time to decide on whether or not this change will be accepted?

PR #249, opened 2023-02-14
TypeScript language support

Theaninova:
resolves #542

This is partially based on the JavaScript implementation, however with significant changes due to how TypeScript can make use of a lot more modern features, like actual classes that will be transpiled in case an older JS version should be targeted.

It might also be worth looking into generating `.js` + `.d.ts` instead...

Output will look something like this
```ts
/*
 * This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild
 */
// @ts-ignore
import KaitaiStream from 'kaitai-struct/KaitaiStream'

export class BitsEnum {
  _is_le?: boolean;

  constructor(
    readonly _io: KaitaiStream,
    readonly _parent: unknown,
    readonly _root: BitsEnum,
  ) {
    this._root ??= this;

    this._read();
  }

  _read() {
    this.one = (this._io.readBitsIntBe(4)) as any
    this.two = (this._io.readBitsIntBe(8)) as any
    this.three = (this._io.readBitsIntBe(1)) as any
  }

  one: BitsEnum.Animal;
  two: BitsEnum.Animal;
  three: BitsEnum.Animal;
}
export namespace BitsEnum {
  export const enum Animal {
    CAT = 0,
    DOG = 1,
    HORSE = 4,
    PLATYPUS = 5,
  }
}
```

To be functional it might also require changes to the JavaScript runtime.

I've tested it against all files in `/test/formats/`, however nothing beyond that.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/Utils.scala
+++ b/shared/src/main/scala/io/kaitai/struct/Utils.scala
@@ -78,6 +78,14 @@ object Utils {
     }
   }
 
+  def lowerKebabCase(s: String): String  = {
+    if (s.startsWith("_")) {
+      "_" + lowerKebabCase(s.substring(1))
+    } else {
+      s.replaceAll("_", "-")
+    }
+  }
+
   /**
     * Converts string to `lowerCamelCase`.
     * @param s original string in `lower_under_score` case.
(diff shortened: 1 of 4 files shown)


Trass3r:
What's the state of this?

PR #254, opened 2023-07-13
Canonicalize encoding names: precompile step, warnings

GreyCat:
This addresses https://github.com/kaitai-io/kaitai_struct/issues/393 at least partially by adding machinery that:

* Introduces list of canonical encoding names (case-sensitive) + aliases (case-insensitive) as EncodingList.scala.
* Checks all values used in `encoding` fields (both in `meta` and in string types).
* If value matches canonical spelling of encoding name precisely, no warning/error issued.
* If value matches canonical spelling case-insensitively, or matches any alias case-insensitively, it fixes encoding to use canonical spelling, and issues a warning.
* If values doesn't match anything, it concedes, leaves it as is, and issues another warning. In some foreseeable future, we'll turn this into an error.

Warning text tests are added in https://github.com/kaitai-io/kaitai_struct_tests/commit/005c2f26121eccd8fb2978bedda46ac8d767d968


shared/src/main/scala/io/kaitai/struct/EncodingList.scala
```diff
@@ -0,0 +1,23 @@
...
+   * Map of encodings supported by Kaitai Struct. Key is the canonical encoding name,
+   * and value is a Set of Strings which comprise all possible aliases of such encoding.
+   * Canonical name itself doesn't need to appear in the list of aliases, so if there's
+   * only one popular spelling of a specific encoding, Set can be empty.
+   *
+   * Canonical names are case-sensitive. Aliases are matched case insensitively and
+   * generate a KSC warning no matter which form was used.
+   */
+  val canonicalToAlias = Map(
+    "US-ASCII" -> Set("ASCII"),
```
generalmimon, line 17: Do we really want to issue warnings and have the compiler suggest to "fix" <code>encoding: <del>ASCII</del></code> (that we've been recommending for years) to <code>encoding: <ins><strong>US-ASCII</strong></ins></code>?

It's not that I wouldn't get used to it, but I don't think anyone would think `ASCII` is "wrong" when it's used much more often in practice.

generalmimon: It's true that for example the standard library of Nim on Windows only recognized `us-ascii` and not `ascii` (see https://github.com/kaitai-io/kaitai_struct/issues/960), so it might make sense to canonicalize to `us-ascii` if it has better support in target languages' standard libs / encoding libraries, but I'm not sure about enforcing `encoding: US-ASCII` in .ksy files (that are written by people) and annoying the users.

generalmimon: > For ASCII, here's what I've got so far

Thanks for the research. This is approximately what I meant by "it might make sense to canonicalize to `us-ascii` if it has better support in target languages' standard libs / encoding libraries", i.e. converting `encoding: ASCII` to `us-ascii` in the generated code in case the encoding library used in a certain target language wouldn't recognize `"ascii"` but recognizes `"us-ascii"` (like the Nim on Windows situation: https://github.com/kaitai-io/kaitai_struct/issues/960). But to decide on what to pick for .ksy specs, I guess it's more important what comes to mind first for an average programmer (to make the experience with Kaitai Struct seamless). The authors of encoding libraries are more likely to work with standards on a daily basis and be a bit biased by that, if you see what I mean.

It may very well be influenced by a programming language, but for example in Python, I've only seen `str.encode('ascii')` so far (e.g. https://docs.python.org/3/howto/unicode.html#converting-to-bytes) and never <code>str.encode('<em>us-ascii</em>')</code>.

GreyCat: It's totally ok to do this conversion when rendering a specific target language based on requirements of that language. I.e. internally we'll use "ASCII" in ksy and everywhere inside the compiler, but NimCompiler will have special mapping table that will render this differently, so everybody will be happy in the end.

PR #255, opened 2023-08-01
Serialization

GreyCat:
Closes https://github.com/kaitai-io/kaitai_struct_compiler/pull/293


shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
```diff
@@ -298,6 +352,39 @@ class ClassCompiler(
...
+
+  def compileSeqWrite(seq: List[AttrSpec], defEndian: Option[FixedEndian]) = {
+    var wasUnaligned = false
+    seq.foreach { (attr) =>
+      val nowUnaligned = isUnalignedBits(attr.dataType)
+      if (wasUnaligned && !nowUnaligned)
+        lang.alignToByte(lang.normalIO)
+      lang.attrWrite(attr, attr.id, defEndian)
+      wasUnaligned = nowUnaligned
+    }
```
GreyCat, line 379: Let's try to avoid using `var` and imperative. `foldLeft` seems to do the trick here, cleanly conveying the semantics and allowing e.g. advanced parallelization of such loops:

```suggestion
    seq.foldLeft(false) { case (wasUnaligned, attr) =>
      val nowUnaligned = isUnalignedBits(attr.dataType)
      if (wasUnaligned && !nowUnaligned)
        lang.alignToByte(lang.normalIO)
      lang.attrWrite(attr, attr.id, defEndian)
      nowUnaligned
    }
```

generalmimon: Actually, one of the major changes in Java and Python runtime libraries related to serialization was that all `alignToByte` logic is handled in the runtime libraries themselves and the `alignToByte()` call generation is disabled in the compiler in both Java and Python:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/2eca3de2c4044b74226d6fcd2f3641a65febd9ab/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala#L497-L500

https://github.com/kaitai-io/kaitai_struct_compiler/blob/2eca3de2c4044b74226d6fcd2f3641a65febd9ab/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L486-L489

(and now I see a small mistake in the comment in `PythonCompiler` - technically it should be "the compiler does not need to output <code><del>alignToByte</del>()</code> <code><ins><strong>align_to_byte</strong></ins>()</code>" per Python naming)

---

So I guess the piece of code you're reviewing can be simplified to just `lang.attrWrite(attr, attr.id, defEndian)`.

shared/src/main/scala/io/kaitai/struct/format/Identifier.scala
```diff
@@ -25,7 +25,7 @@ abstract class Identifier {
   */
 case class NumberedIdentifier(idx: Int) extends Identifier {
   import NumberedIdentifier._
-  override def humanReadable: String = s"${TEMPLATE}_$idx"
+  override def humanReadable: String = s"_${TEMPLATE}$idx"
```
GreyCat, line 28: Out of curiosity, why do we need to change this?

shared/src/main/scala/io/kaitai/struct/datatype/Endianness.scala
```diff
@@ -57,7 +57,7 @@ object Endianness {
   def fromString(s: Option[String], defaultEndian: Option[Endianness], dt: String, path: List[String]): Option[FixedEndian] = s match {
     case Some("le") => Some(LittleEndian)
     case Some("be") => Some(BigEndian)
-    case None =>
+    case _ =>
```
GreyCat, line 60: Does  this change endianness parsing behavior?

shared/src/main/scala/io/kaitai/struct/format/Identifier.scala
generalmimon: > why do we need to change this?

We don't - as the name suggests, this is exclusively a "human-readable" identifier, so it sometimes shows up in error messages, but it shouldn't affect anything functional.

shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
```diff
@@ -145,12 +171,82 @@ class PythonCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
       out.puts("pass")
   }
 
-  override def readFooter() = universalFooter
+  override def fetchInstancesHeader(): Unit = {
+    out.puts
+    out.puts("def _fetch_instances(self):")
+    out.inc
+    out.puts("pass")
```
Brachi, line 178: Is there a limitation to make the insertion of `pass` conditional? Like in [readHeader](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L170) with `isEmpty`. For this particular function it doesn't look so straightforward given that [attrs](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala#L326) seems to always be non-empty.

Although innocuous, having a "dangling" `pass` stands out on first read and makes one wonder if there was some kind of bug in the code generation or in the source ksy file. Probably a nit though!

```
def _fetch_instances(self):
    pass
    self.attr_01._fetch_instances()
    ...
```

Same occurrences  in: 
- [checkHeader](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L213)
- [checkInstanceHeader](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L227)
- [condIfHeader](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L530)
- [condRepeatCommonHeader](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L582)
- [switchIfCaseFirstStart](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L680)
- [switchIfCaseStart](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L686)
- [switchIfElseStart](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L695)

generalmimon: > Is there a limitation to make the insertion of `pass` conditional? Like in [readHeader](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L170) with `isEmpty`. For this particular function it doesn't look so straightforward

Yep, it's not very straightforward with how the current compiler code is internally structured. The problem is that when the compiler generates a method header (typically in the `*Header` methods), it doesn't know whether that function will end up having some body or not. In case of `readHeader`, we're just very lucky that the body will be empty if and only if there are `seq` fields. Things get a bit more complicated in case of `writeHeader`, but it's doable - generated `_write` methods might not be empty even if there are no `seq` fields, provided there are parse instances that need their "write itself upon access" flags initialized:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/cb0c1ebfa87e7eb8b86756be689e4840cd776544/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala#L333

For `_check` or `_fetchInstances` (let alone something reused for _many_ different purposes like `condIfHeader`!), trying to predict ahead of time with a ultimate boolean expression whether the function will or won't be empty borders on insanity (again, with the current code structure). It's of course theoretically possible, just absolutely unmaintainable and definitely not worth spending time on.

I'll rather take all the superfluous `pass`es - the worst thing that can happen is that someone inspects the generated code and the `pass`es "stand out" to them, as you put it. Far worse would be if we tried to eliminate them, but in that effort we forget to insert a `pass` in some cases when it should be there, which means the entire generated Python code will be rejected by Python interpreter and people will actually have to patch these manually because KS compiler produced invalid code.

So until there's time and willingness to refactor the compiler code so that we can reliably know whether the function body is empty or not at the time of generating its header, I don't think there's any point in doing anything about this. And this is very low priority for me, given that it only affects code purity. Almost every KS issue should have a higher priority than this.

Brachi: Thanks a lot for the long and insightful answer! This confirms my suspicion that it was tricky. I agree this is very low priority, sorry for the noise\[1\].
Thanks also for the tremendous effort done in serialization. I'm super excited about this feature :D

\[1\] At the risk of adding more noise, just in case I wanted to point out that an empty docstring is an alternative to using `pass` for an empty body function. But probably without content it's still not worth it.

generalmimon:
@GreyCat Could you please continue with the review?

shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
generalmimon: For completeness: I've realized it wouldn't be hard to only insert `pass`es when they're needed if we added `pass` in the `*Footer` method instead of the `*Header` method. At the time of when `*Footer` is called, we already know whether there was some code inserted since the corresponding `*Header`.

dakhnod:
@GreyCat @generalmimon Hey, how is this getting along? Would be great to have serialization support in main!

jhgorse:
Looks like we are 359 commits behind master. It could use a rebase. I tried running the `sbt test` to smoke test the build but the tests fail in `serialization` branch. There are also 89 merge conflicts to master.

generalmimon:
@jhgorse:

> Looks like we are 359 commits behind master.

> There are also 89 merge conflicts to master.

Don't worry, this will be in 0.11 no matter what - see also https://github.com/kaitai-io/kaitai_struct/issues/1060. I just don't think it's any less work to resolve conflicts regularly (like every week) than to resolve them all at once.

> I tried running the `sbt test` to smoke test the build but the tests fail in `serialization` branch.

This is normal and has also been reported in https://github.com/kaitai-io/kaitai_struct/issues/852. In short, the reason is that we also use tests to track known bugs.  You can also see this at https://ci.kaitai.io/ - there is no language which passes 100% tests.

If any software passes 100% of the tests, then it is either bug-free (which is extremely unlikely) or the test suite is insufficient.

PR #257, opened 2023-08-25
Implement classToString for Nim

sealmove:
@GreyCat Please confirm this looks good. This is an example of how it works in Nim https://play.nim-lang.org/#ix=4EtZ.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
@@ -45,6 +45,15 @@ class NimCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     out.puts
   }
 
+  override def classToString(toStringExpr: Ast.expr): Unit = {
+    out.puts
+    out.puts(s"proc `$$`(x: ${namespaced(typeProvider.nowClass.name)}): string =")
+    out.inc
+    out.puts(s"return ${translator.translate(toStringExpr)}")
+    out.dec
+    out.puts
+  }
+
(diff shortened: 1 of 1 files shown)


GreyCat:
Looks good, let's merge this one!

shared/src/main/scala/io/kaitai/struct/languages/NimCompiler.scala
```diff
@@ -45,6 +45,15 @@ class NimCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
     out.puts
   }
 
+  override def classToString(toStringExpr: Ast.expr): Unit = {
+    out.puts
+    out.puts(s"proc `$$`(x: ${namespaced(typeProvider.nowClass.name)}): string =")
+    out.inc
+    out.puts(s"return ${translator.translate(toStringExpr)}")
```
generalmimon, line 52: I'm curious how the `toStringExpr` expression will know that it should prefix all attribute references with `x.`?

generalmimon:
It's nice that this PR implements the `classToString` method in `NimCompiler`, but it's not used from anywhere. Nim uses a custom [`NimClassCompiler`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/ae16deed271620bf8f06b29894e67a6c7350a969/shared/src/main/scala/io/kaitai/struct/NimClassCompiler.scala) class, which (unlike `ClassCompiler` - see [`ClassCompiler.scala:105`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/ae16deed271620bf8f06b29894e67a6c7350a969/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala#L105)) doesn't call `classToString`.

This is what we get for copying code, BTW.

PR #230, opened 2020-12-05
Upgrade all dependencies to the latest versions

Mingun:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/683

PR #258, opened 2023-10-14
Basic f-strings (intepolated strings)

GreyCat:
Implements step 1 out of [formatting strings proposal](https://github.com/kaitai-io/kaitai_struct/issues/1073).

* Added basic f-string parsing in expression language (step 1 only, just interpolation, without any extra formatting specification)
* Added basic interpolated string (f-string) translation into target languages, mostly working off concatenation and existing other-types-to-string conversions. Supports only integers and strings now.
  * GoTranslator: given non-string nature, added custom implementation using `fmt.Sprintf`
  * CommonLiterals: split generation of string "body" (without quotes) and adding quotes to it
  * CommonMethods: param name cleanup

Both are supported by unit tests.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
@@ -261,6 +261,10 @@ class ExpressionsSpec extends AnyFunSpec {
       Expressions.parse("\"abc\\u21bbdef\"") should be (Str("abc\u21bbdef"))
     }
 
+    it("parses double-quoted string with double quote") {
+      Expressions.parse("\"this \\\" is a quote\"") should be(Str("this \" is a quote"))
+    }
+
     // Casts
     it("parses 123.as<u4>") {
       Expressions.parse("123.as<u4>") should be (
@@ -387,6 +391,68 @@ class ExpressionsSpec extends AnyFunSpec {
 
     it("parses foo.bar") {
       Expressions.parse("foo.bar") should be (Attribute(Name(identifier("foo")),identifier("bar")))
+    }
+
+    describe("f-strings") {
+      it("parses f-string with just a string") {
+        Expressions.parse("f\"abc\"") should be(InterpolatedStr(ArrayBuffer(
+          Str("abc")
+        )))
+      }
+
+      it("parses f-string with just one expression") {
+        Expressions.parse("f\"{123}\"") should be(InterpolatedStr(ArrayBuffer(
+          IntNum(123)
+        )))
+      }
+
+      it("parses f-string with string + expression") {
+        Expressions.parse("f\"foo={123}\"") should be(InterpolatedStr(ArrayBuffer(
+          Str("foo="),
+          IntNum(123)
+        )))
+      }
+
+      it("parses f-string with expression + string") {
+        Expressions.parse("f\"{123}=abc\"") should be(InterpolatedStr(ArrayBuffer(
+          IntNum(123),
+          Str("=abc")
+        )))
+      }
+
+      it("parses f-string with str + expression + str") {
+        Expressions.parse("f\"abc={123}=def\"") should be(InterpolatedStr(ArrayBuffer(
+          Str("abc="),
+          IntNum(123),
+          Str("=def")
+        )))
+      }
+
+      it("parses f-string string with newline in the middle") {
+        Expressions.parse("f\"abc\\ndef\"") should be(InterpolatedStr(ArrayBuffer(Str("abc\ndef"))))
+      }
+
+      it("parses f-string with double quote in the middle") {
+        Expressions.parse("f\"this \\\" is a quote\"") should be(InterpolatedStr(ArrayBuffer(
+          Str("this \" is a quote")
+        )))
+      }
+
+      it("parses f-string with string in it") {
+        Expressions.parse("f\"abc{\"def\"}ghi\"") should be(InterpolatedStr(ArrayBuffer(
+          Str("abc"),
+          Str("def"),
+          Str("ghi"),
+        )))
+      }
+
+      it("parses f-string with f-string in it") {
+        Expressions.parse("f\"abc{f\"def\"}ghi\"") should be(InterpolatedStr(ArrayBuffer(
+          Str("abc"),
+          InterpolatedStr(ArrayBuffer(Str("def"))),
+          Str("ghi"),
+        )))
+      }
     }
   }
 }
--- a/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
@@ -395,33 +395,33 @@ class ExpressionsSpec extends AnyFunSpec {
 
     describe("f-strings") {
       it("parses f-string with just a string") {
-        Expressions.parse("f\"abc\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"abc\"") should be(InterpolatedStr(Seq(
           Str("abc")
         )))
       }
 
       it("parses f-string with just one expression") {
-        Expressions.parse("f\"{123}\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"{123}\"") should be(InterpolatedStr(Seq(
           IntNum(123)
         )))
       }
 
       it("parses f-string with string + expression") {
-        Expressions.parse("f\"foo={123}\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"foo={123}\"") should be(InterpolatedStr(Seq(
           Str("foo="),
           IntNum(123)
         )))
       }
 
       it("parses f-string with expression + string") {
-        Expressions.parse("f\"{123}=abc\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"{123}=abc\"") should be(InterpolatedStr(Seq(
           IntNum(123),
           Str("=abc")
         )))
       }
 
       it("parses f-string with str + expression + str") {
-        Expressions.parse("f\"abc={123}=def\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"abc={123}=def\"") should be(InterpolatedStr(Seq(
           Str("abc="),
           IntNum(123),
           Str("=def")
@@ -429,17 +429,17 @@ class ExpressionsSpec extends AnyFunSpec {
       }
 
       it("parses f-string string with newline in the middle") {
-        Expressions.parse("f\"abc\\ndef\"") should be(InterpolatedStr(ArrayBuffer(Str("abc\ndef"))))
+        Expressions.parse("f\"abc\\ndef\"") should be(InterpolatedStr(Seq(Str("abc\ndef"))))
       }
 
       it("parses f-string with double quote in the middle") {
-        Expressions.parse("f\"this \\\" is a quote\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"this \\\" is a quote\"") should be(InterpolatedStr(Seq(
           Str("this \" is a quote")
         )))
       }
 
       it("parses f-string with string in it") {
-        Expressions.parse("f\"abc{\"def\"}ghi\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"abc{\"def\"}ghi\"") should be(InterpolatedStr(Seq(
           Str("abc"),
           Str("def"),
           Str("ghi"),
@@ -447,9 +447,9 @@ class ExpressionsSpec extends AnyFunSpec {
       }
 
       it("parses f-string with f-string in it") {
-        Expressions.parse("f\"abc{f\"def\"}ghi\"") should be(InterpolatedStr(ArrayBuffer(
+        Expressions.parse("f\"abc{f\"def\"}ghi\"") should be(InterpolatedStr(Seq(
           Str("abc"),
-          InterpolatedStr(ArrayBuffer(Str("def"))),
+          InterpolatedStr(Seq(Str("def"))),
           Str("ghi"),
         )))
       }
--- a/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
@@ -189,80 +189,91 @@ class ExpressionsSpec extends AnyFunSpec {
       )
     }
 
-    // Boolean literals
-    it("parses true") {
-      Expressions.parse("true") should be (Bool(true))
+    describe("boolean literals") {
+      it("parses true") {
+        Expressions.parse("true") should be(Bool(true))
+      }
+
+      it("parses false") {
+        Expressions.parse("false") should be(Bool(false))
+      }
+
+      it("parses truer") {
+        Expressions.parse("truer") should be(Name(identifier("truer")))
+      }
     }
 
-    it("parses false") {
-      Expressions.parse("false") should be (Bool(false))
-    }
-
-    it("parses truer") {
-      Expressions.parse("truer") should be (Name(identifier("truer")))
-    }
-
-    // Boolean operations
-    it("parses not foo") {
-      Expressions.parse("not foo") should be (
-        UnaryOp(
-          Ast.unaryop.Not,
-          Name(identifier("foo"))
-        )
-      )
-    }
-
-    it("parses note_len") {
-      Expressions.parse("note_len") should be (Name(identifier("note_len")))
-    }
-
-    it("parses notnot") {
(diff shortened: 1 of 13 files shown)


shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -469,6 +471,35 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
...
+      case _ =>
+        importList.add("fmt")
+
+        val piecesAndArgs: Seq[(String, Option[String])] = exprs.map {
+          case Ast.expr.Str(s) => (doStringLiteralBody(s), None)
+          case e =>
+            detectType(e) match {
+              case _: IntType => ("%d", Some(translate(e)))
+              case _: StrType => ("%s", Some(translate(e)))
+              case _: BooleanType => ("%b", Some(translate(e)))
```
generalmimon, line 490: See https://go.dev/play/p/QyIvwJ5P8fZ:

```go
package main

import "fmt"

func main() {
	fmt.Printf("%b\n", true)
}
```

```
./prog.go:6:2: fmt.Printf format %b has arg true of wrong type bool

Go vet failed.
```

Note that `%b` means "binary" formatting, so I don't know what you meant here. See https://pkg.go.dev/fmt#hdr-Printing:

> Integer:
> 
> ```
> %b	base 2
> ```
> 
> Floating-point and complex constituents:
> 
> ```
> %b	decimalless scientific notation with exponent a power of two,
> 	in the manner of strconv.FormatFloat with the 'b' format,
> 	e.g. -123456p-78
> ```

---

@GreyCat I wonder why you don't just use `%v` by default for everything?

> General:
> ```
> %v	the value in a default format
> 		when printing structs, the plus flag (%+v) adds field names
> ```

This program:

```go
package main

import "fmt"

func main() {
	fmt.Printf("%v\n", true)
}
```

has the following output:

```
true

Program exited.
```

You can also use `%t` which works exclusively for booleans, but I don't really see the benefit over `%v`:

> Boolean:
> 
> ```
> %t	the word true or false
> ```

See also what the universal `%v` boils down to for individual types:

> The default format for %v is:
> 
> ```
> bool:                    %t
> int, int8 etc.:          %d
> uint, uint8 etc.:        %d, %#x if printed with %#v
> float32, complex64, etc: %g
> string:                  %s
> chan:                    %p
> pointer:                 %p
> ```

shared/src/main/scala/io/kaitai/struct/translators/CommonLiterals.scala
```diff
@@ -8,16 +8,27 @@ trait CommonLiterals {
...
+  /**
+   * Generates string literal with double quotes framing it.
+   * @param s string to put in as literal
+   * @return string literal
+   */
+  def doStringLiteral(s: String): String =
+    "\"" + doStringLiteralBody(s) + "\""
+
+  /**
+   * Generates body of string literal for a given string, without framing quotes.
```
generalmimon, line 20:
```suggestion
   * Generates body of string literal for a given string, without enclosing quotes.
```

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -469,6 +471,35 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
...
+    exprs match {
+      case Seq(Ast.expr.Str(s)) =>
+        // exactly one string literal, no need for printf at all
+        trStringLiteral(s)
+
+      case _ =>
+        importList.add("fmt")
+
+        val piecesAndArgs: Seq[(String, Option[String])] = exprs.map {
+          case Ast.expr.Str(s) => (doStringLiteralBody(s), None)
```
generalmimon, line 485: Given that anything you produce here goes directly to the format string of [`fmt.Sprintf`](https://pkg.go.dev/fmt#Sprintf), you have to escape any potential percent sign `%`, see https://pkg.go.dev/fmt#hdr-Printing:

> General:
> 
> ```
> %%	a literal percent sign; consumes no value
> ```

Also please add a test for this in `TranslatorSpec`.

jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
```diff
@@ -388,5 +388,41 @@ class ExpressionsSpec extends AnyFunSpec {
     it("parses foo.bar") {
       Expressions.parse("foo.bar") should be (Attribute(Name(identifier("foo")),identifier("bar")))
     }
+
+    describe("f-strings") {
```
generalmimon, line 392: I'd also add to have a test for:

1. a literal quote `"` inside an f-string, i.e. something like `Expressions.parse("f\"this \\\" is a quote\"")`
2. a normal string inside an f-string: `Expressions.parse("f\"here we {\"have\"} a normal string\"")`
3. an f-string inside an f-string, for example: `Expressions.parse("f\"here we {f\"have\"} another f-string\"")`

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -469,6 +471,35 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
     ResultLocalVar(v)
   }
 
+  def trInterpolatedStringLiteral(exprs: Seq[Ast.expr]): TranslatorResult = {
+
+    exprs match {
```
generalmimon, line 476: I don't like the extra newline here, but that's absolutely minor:

```suggestion
  def trInterpolatedStringLiteral(exprs: Seq[Ast.expr]): TranslatorResult = {
    exprs match {
```

shared/src/main/scala/io/kaitai/struct/translators/CommonLiterals.scala
```diff
@@ -8,16 +8,27 @@ trait CommonLiterals {
...
-      if (code <= 0xff) {
-        strLiteralAsciiChar(code)
-      } else {
-        strLiteralUnicode(code)
-      }
-    ).mkString
-    "\"" + encoded + "\""
-  }
+  /**
+   * Generates string literal with double quotes framing it.
```
generalmimon, line 12:
```suggestion
   * Generates string literal enclosed in double quotes.
```

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
```diff
@@ -204,4 +206,19 @@ abstract class BaseTranslator(val provider: TypeProvider)
...
+    exprs.map(anyToStr).mkString(" + ")
+
+  def anyToStr(value: Ast.expr): String = {
+    detectType(value) match {
+      case _: IntType =>
+        intToStr(value, Ast.expr.IntNum(10))
+      case _: StrType =>
+        translate(value)
+      case otherType =>
+        throw new UnsupportedOperationException(s"unable to convert $otherType to string in format string")
```
generalmimon, line 221:
```suggestion
        throw new UnsupportedOperationException(s"unable to convert $otherType to string in format string (only integers and strings are supported)")
```

generalmimon (changes requested):

generalmimon:
@GreyCat I also have a question - is there a way to get a literal `{` in the f-string? From what I see in the expression parsing code, it doesn't look like there is, but I may have missed something.

In Python, for example, you can escape a `{` by doubling it - I wonder if we can get this too?

```py
>>> f"1 + 2 = {1 + 2} {{basic math}}"
'1 + 2 = 3 {basic math}'
```

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
GreyCat: That's clearly not intentional, thanks for catching this!

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
GreyCat: I love your reviews! Thanks for catching!

I plan to add .kst tests for this feature (to test this end-to-end — making sure that generated target language code actually works) and I think it will be better covered by .kst tests.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
GreyCat: Again, thanks for catching! I indeed made an assumption and was proven wrong.

`%v` is a good idea, but I wonder if we'll still need to get back to individual `%d` / `%g` / `%s` / `%t`, etc, once we'll get to different formatting options. If we'll get request for hex formatting, we'll still need to do `%x`, etc. If we'll be looking into zero-padding or field width, it will be also different for different types.

Anyway, let be fix it first.

jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
GreyCat: Ok, the very first proposed test of yours actually pinpointed a bug that I've already spent several hours trying to get my head around and it doesn't work ;)

Basically, the test: `Expressions.parse("f\"this \\\" is a quote\"")`

```
Expected :InterpolatedStr(ArrayBuffer(Str(this " is a quote)))
Actual   :InterpolatedStr(ArrayBuffer(Str(this "is a quote)))
```

So, the current parser eats the whitespace *after* the escaped char. It seems to have something to do with implicit whitespace handling that our parsed ripped from FastParse's Python parser seem to employ... No idea though how to dig deeper into this...

generalmimon: Maybe https://com-lihaoyi.github.io/fastparse/#WhitespaceHandling can help?

> The whitespace consumer affects the `~` and `.rep` operators to consume all non-trailing whitespace and ignoring it (**use `~~` and `.repX` instead if you need to access non-whitespace-consuming operators in certain cases**).

Unfortunately, it's for FastParse 3.0.2, while we're still stuck with FastParse 1.0.0 (so I don't know if that works also for the version we're using).

I've been eyeing https://github.com/kaitai-io/kaitai_struct_compiler/pull/230 for a while now (which upgrades to FastParse 2.3.3), but unfortunately, it's based on https://github.com/kaitai-io/kaitai_struct_compiler/pull/222 (see https://github.com/kaitai-io/kaitai_struct_compiler/pull/230#issuecomment-840791757) and I know that if I merge https://github.com/kaitai-io/kaitai_struct_compiler/pull/222, there's an issue in https://github.com/kaitai-io/kaitai_struct_webide with no longer working TypeScript definitions (which go a bit over my head, but admittedly I haven't spent that much time on this).

https://web.archive.org/web/20181016164455/http://www.lihaoyi.com/fastparse/#WhitespaceHandling claims to be for FastParse 1.0.0 that we use, but it doesn't seem to contain the part about the non-whitespace-consuming operators.

generalmimon: It seems to me that the problem in our case is the `.rep` on `fstringElement`:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/17f9e40a8d81c767e6a4f648cfad2d4d66bb68e1/shared/src/main/scala/io/kaitai/struct/exprlang/Expressions.scala#L38

```scala
  val fstring: P[Ast.expr.InterpolatedStr] = P("f\"" ~/ fstringElement.rep ~ "\"").map(Ast.expr.InterpolatedStr)
  //                                                                  ^^^^
```

generalmimon: @GreyCat 

Changing `rep` to `repX` here fixes the "a quote in the middle" case:

```diff
diff --git i/shared/src/main/scala/io/kaitai/struct/exprlang/Expressions.scala w/shared/src/main/scala/io/kaitai/struct/exprlang/Expressions.scala
index d39a157e..dd915e1b 100644
--- i/shared/src/main/scala/io/kaitai/struct/exprlang/Expressions.scala
+++ w/shared/src/main/scala/io/kaitai/struct/exprlang/Expressions.scala
@@ -38,7 +38,7 @@ object Expressions {
   val fstring: P[Ast.expr.InterpolatedStr] = P("f\"" ~/ fstringElement.rep ~ "\"").map(Ast.expr.InterpolatedStr)
   val fstringElement: P[Ast.expr] = P(
     formatExpr |
-    fstringItem.rep(min = 1).map(_.mkString).map(Ast.expr.Str)
+    fstringItem.repX(min = 1).map(_.mkString).map(Ast.expr.Str)
   )
   val fstringItem = P(fstringChar.! | Lexical.escapeseq)
   val fstringChar = P(CharsWhile(!"{\\\"".contains(_)))
```

Meaning this test now passes:

```diff
-      it("parses f-string with double quote in it") {
+      it("parses f-string with double quote in the middle") {
         Expressions.parse("f\"this \\\" is a quote\"") should be(InterpolatedStr(ArrayBuffer(
           Str("this \" is a quote")
         )))
       }
```

But this one still doesn't:

```scala
      it("parses f-string with double quote at the start") {
        Expressions.parse("f\" \\\" is a quote\"") should be(InterpolatedStr(ArrayBuffer(
          Str(" \" is a quote")
        )))
      }
```

```
[info] ExpressionsSpec:
[info]   - parses f-string with double quote at the start *** FAILED ***
[info]     InterpolatedStr(ArrayBuffer(Str(" is a quote))) was not equal to InterpolatedStr(ArrayBuffer(Str( " is a quote))) (ExpressionsSpec.scala:438)
[info]     Analysis:
[info]     Ast$expr$InterpolatedStr(elts: ArrayBuffer(0: Str(" is a quote) -> Str( " is a quote)))
```

In summary, we'll need a little more tests to catch bugs in all places :)

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
GreyCat: Fixed.

GreyCat:
> @GreyCat I also have a question - is there a way to get a literal `{` in the f-string? From what I see in the expression parsing code, it doesn't look like there is, but I may have missed something.
> 
> In Python, for example, you can escape a `{` by doubling it - I wonder if we can get this too?

To be honest, I don't want to overcomplicate it, nor I see a lot of value in having this escaping behavior.

In Python itself, it looks pretty weird per se — e.g.:

* `{}` is a valid expression to specify empty dict.
* `{foo}` is a valid expression for one-element set containing whatever `foo` variable has
* `{foo:x}` is a valid expression for a dict containing one key-value pair; key is whatever `foo` variable has, and value is whatever `x` variable has
 
Therefore, if I want to include a set in f-string, I'm down to unpleasant surprise:

```
>>> foo = "bar"
>>> x = 42
>>> f"{{foo}}"
'{foo}'
>>> f"{{foo:x}}"
'{foo:x}'
```

It's possible to work around with extra spaces:

```
>>> f"{ {foo}}"
"{'bar'}" # as expected
>>> f"{ {foo} }"
"{'bar'}" # as expected
>>> f"{ {foo:x}}"
"{'bar': 42}"
>>> f"{ {foo:x} }"
"{'bar': 42}"
```

... and then you try some other combinations and it gets progressively weirder, e.g. this throws a very weird error:

```
>>> f"{{}:x}"
  File "<stdin>", line 1
    f"{{}:x}"
             ^
SyntaxError: f-string: single '}' is not allowed
```

while adding initial space makes it work as expected:

```
>>> f"{ {}:x}"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported format string passed to dict.__format__
```

So, my proposal is basically avoiding going into that rabbit hole, at least yet. If anybody wants to have literal `{` or `}` in our f-strings, just keep it simple and do `f"1 + 2 = {1 + 2} " + "{basic math}"`

jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
GreyCat: Ok, should be good now, I think I've figured out that using `~~` (or `~~/`) operator + `repX` instead of `rep` does the trick of bypassing this whitespace eating implicits.

Added the test you've suggested as well as couple others (e.g. to test whitespace at the end of the string), should be good now.

shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
GreyCat: Fixed that, updated main TranslatorSpec test to reflect this situation. Will still do .kst test in later stages of this project.

GreyCat:
@generalmimon If you'll have some time, can I ask you to take another look? It looks like I've accommodated all your feedback, implemented all the fixes, and rebased it off the latest master — so it looks reasonably good as a first step to merge in.

Meanwhile I'll start adding .kst tests for this and once this is good to go, I'll start stage 2, which will hopefully give us better formatting options. Then, finally, we can probably deprecate `-web-ide-representation`.

GreyCat:
Interpreting thumbs up as all good — therefore merging. Thanks!

PR #265, opened 2024-02-26
CanonicalizeEncodingNames: fix locale-sensitive `toUpperCase`

generalmimon:
Avoid the problem with the Turkish locale that we have encountered in the past, see https://github.com/kaitai-io/kaitai_struct/issues/708.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames$Test.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames$Test.scala
@@ -32,6 +32,8 @@ class CanonicalizeEncodingNames$Test extends AnyFunSpec {
     }
 
     it("reports warning and fixes bad capitalization for 'iSo-8859-1' even in Turkish locale") {
+      // This test only covers the case conversion in the `canonicalizeName` implementation,
+      // not the case conversions used when initializing the `aliasToCanonical` map.
       val oldLocale = Locale.getDefault
       Locale.setDefault(new Locale("tr"))
       try {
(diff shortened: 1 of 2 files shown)


jvm/src/test/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames$Test.scala
```diff
@@ -29,5 +30,17 @@ class CanonicalizeEncodingNames$Test extends AnyFunSpec {
       newEncoding should be("ISO-8859-1")
       problem should be(Some(EncodingNameWarning("ISO-8859-1", "iSo-8859-1")))
     }
+
+    it("reports warning and fixes bad capitalization for 'iSo-8859-1' even in Turkish locale") {
+      val oldLocale = Locale.getDefault
+      Locale.setDefault(new Locale("tr"))
+      try {
+        val (newEncoding, problem) = CanonicalizeEncodingNames.canonicalizeName("iSo-8859-1")
```
generalmimon, line 38: Unfortunately, this only tests the case conversion directly in `CanonicalizeEncodingNames.canonicalizeName`:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/4ddd3c70a2aa3e6d5ea1f0a76a849d7fd7ccfa0a/shared/src/main/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames.scala#L52

Meaning that these occurrences are untested:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/4ddd3c70a2aa3e6d5ea1f0a76a849d7fd7ccfa0a/shared/src/main/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames.scala#L67-L71

I presume this `aliasToCanonical` initialization runs at the time the `object CanonicalizeEncodingNames` is loaded, so it's probably not so simple to change the default locale & re-run it dynamically, as I did for `canonicalizeName`. Perhaps there's a way to do that using [`ClassLoader`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html) or something similar, but I have no idea.

GreyCat: Yeah, I'm leaning towards it doesn't worth the complexity for the test. It looks like it works if tested manually, so seems to be good to go.

GreyCat: Maybe worth adding this remark as a comment to the test implementation, though.

GreyCat (approved):

PR #269, opened 2024-03-06
CommonMethods: declarative specification of methods

GreyCat:
* Refactored CommonMethods to use declarative specification of methods,  and build all checks for arguments quantity and types around that, equally for all the data types and implementations.
  * Introduced set of new case objects derived from `MethodArgType` to be able to embody expectations about argument types (e.g. `case _: IntType` will be represented as `IntArg`).
  * Introduced set of new `MethodSig` classes to be able to declaratively specify method signatures.
* Added `ExpressionValidator$Test` that will test that functionality (as ExpressionValidator will be first in line, resuing CommonMethods functionality)
* Added a few new exceptions to clearly communicate CommonMethods checks results (WrongMethodCall, MethodNotFoundErrorWithArg).
* ExpressionValidator: fixed problem with validation of subscripts usage (and added relevant test to `ExpressionValidator$Test`).
* Fixed `enum` warning -> `enumName` for Scala 3 compatibility.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator$Test.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator$Test.scala
@@ -1,0 +1,204 @@
+package io.kaitai.struct.translators
+
+import io.kaitai.struct.datatype.DataType._
+import io.kaitai.struct.exprlang.Expressions
+import io.kaitai.struct.exprlang.Expressions.ParseException
+import io.kaitai.struct.precompile.{MethodNotFoundErrorWithArg, TypeMismatchError, WrongMethodCall}
+import io.kaitai.struct.translators.TestTypeProviders._
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers._
+
+class ExpressionValidator$Test extends AnyFunSpec {
+  val alwaysInt = Always(CalcIntType)
+  val alwaysIntValidator = new ExpressionValidator(alwaysInt)
+
+  describe("simple literals") {
+    describe("valid") {
+      it("123") {
+        val ex = Expressions.parse("123")
+        alwaysIntValidator.validate(ex)
+      }
+
+      it("123.456e12") {
+        val ex = Expressions.parse("123.456e12")
+        alwaysIntValidator.validate(ex)
+      }
+
+      it("\"foo\"") {
+        val ex = Expressions.parse("\"foo\"")
+        alwaysIntValidator.validate(ex)
+      }
+    }
+
+    describe("broken") {
+      it("123ff") {
+        val thrown = the [ParseException] thrownBy Expressions.parse("123ff")
+        thrown.getMessage should be("Position 1:4, found \"ff\"")
+      }
+
+      it("123.456e12ff") {
+        val thrown = the [ParseException] thrownBy Expressions.parse("123.456e12ff")
+        thrown.getMessage should be("Position 1:11, found \"ff\"")
+      }
+
+      it("\"foo") {
+        val thrown = the [ParseException] thrownBy Expressions.parse("\"foo")
+        thrown.getMessage should be("Position 1:5, found \"\"")
+      }
+    }
+  }
+
+  describe("integer methods") {
+    describe("valid") {
+      it("123.to_s") {
+        val ex = Expressions.parse("123.to_s")
+        alwaysIntValidator.validate(ex)
+      }
+
+      it("123.to_s()") {
+        val ex = Expressions.parse("123.to_s()")
+        alwaysIntValidator.validate(ex)
+      }
+    }
+
+    describe("broken") {
+      it("123.to_s(3)") {
+        val ex = Expressions.parse("123.to_s(3)")
+        val thrown = the[WrongMethodCall] thrownBy alwaysIntValidator.validate(ex)
+        thrown.getMessage should be("wrong arguments to method call `to_s` on integer: expected (), got (IntNum(3))")
+      }
+
+      it("123.unknown_method") {
+        val ex = Expressions.parse("123.unknown_method")
+        val thrown = the[MethodNotFoundErrorWithArg] thrownBy alwaysIntValidator.validate(ex)
+        thrown.getMessage should be("don't know how to call method 'unknown_method' of object type 'integer'")
+      }
+
+      it("123.unknown_method_with_param(true)") {
+        val ex = Expressions.parse("123.unknown_method_with_param(true)")
+        val thrown = the[MethodNotFoundErrorWithArg] thrownBy alwaysIntValidator.validate(ex)
+        thrown.getMessage should be("don't know how to call method 'unknown_method_with_param' of object type 'integer'")
+      }
+    }
+  }
+
+  describe("float methods") {
+    describe("valid") {
+      it("1.234.to_i") {
+        val ex = Expressions.parse("1.234.to_i")
+        alwaysIntValidator.validate(ex)
+      }
+    }
+
+    describe("broken") {
+      it("1.234.unknown_method") {
+        val ex = Expressions.parse("1.234.unknown_method")
+        val thrown = the[MethodNotFoundErrorWithArg] thrownBy alwaysIntValidator.validate(ex)
+        thrown.getMessage should be("don't know how to call method 'unknown_method' of object type 'float'")
+      }
+
+      it("1.234.unknown_method_with_param(true)") {
+        val ex = Expressions.parse("1.234.unknown_method_with_param(true)")
+        val thrown = the[MethodNotFoundErrorWithArg] thrownBy alwaysIntValidator.validate(ex)
+        thrown.getMessage should be("don't know how to call method 'unknown_method_with_param' of object type 'float'")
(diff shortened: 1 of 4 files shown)


generalmimon:
To begin with, I just have to make one technical remark: please resist writing "nameless" commit messages like the one in https://github.com/kaitai-io/kaitai_struct_compiler/pull/269/commits/580e4bc0d2ec229cfdceb9c19ba69a733fb6467f. I mean the detail of the content is fine, but the problem is that it's missing the title line, so Git tools don't have a good way to display this. GitHub truncates it after the first line break (which means that I only get to see the beginning of your first bullet point), but my usual way of displaying the commit tree locally doesn't, which is arguably even worse:

image

Needless to say, if you feel the urge to cram so many bullet points into one commit, it should've likely been multiple commits. But you probably know this ;)

---

@GreyCat I strongly recommend you to check out https://cbea.ms/git-commit/#seven-rules. The relevant and the most important rule in my opinion is https://cbea.ms/git-commit/#separate - "Separate subject from body with a blank line".

Another good rule is to keep the subject line short, ideally under 50 characters - https://cbea.ms/git-commit/#limit-50. I personally try to follow this, so I know very well that 50 characters is often not much for what I want to tell the world, so I often allow myself to go with 65 or so. Just like they recommend:

> So shoot for 50 characters, but consider 72 the hard limit.

shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
```diff
@@ -4,9 +4,137 @@ import io.kaitai.struct.datatype.DataType
...
+    override def toString = "float"
+  }
+  case object StrArg extends MethodArgType {
+    override def toString = "string"
+  }
+  case object BooleanArg extends MethodArgType {
+    override def toString = "boolean"
+  }
+  case object BytesArg extends MethodArgType {
+    override def toString = "bytes array"
```
generalmimon, line 24:
```suggestion
    override def toString = "byte array"
```

shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
```diff
@@ -4,9 +4,137 @@ import io.kaitai.struct.datatype.DataType
...
+      case _: StrType => Some(StrArg)
+      case _: BooleanType => Some(BooleanArg)
+      case _: BytesType => Some(BytesArg)
+      case _: ArrayType => Some(ArrayArg)
+      case _ => None
+    }
+  }
+
+  def isArgAcceptable(actualType: DataType, expectedType: MethodArgType) =
+    byDataType(actualType).getOrElse(false) == expectedType
```
generalmimon, line 43: I'm a bit shocked that this type mismatch between left and right side of `==` (left side `MethodArgType | Boolean`, right side `MethodArgType`) isn't a compile-time error, but Scala apparently doesn't see a problem in testing different types for equality.

Anyway, I'd rather see this:

```suggestion
  def isArgAcceptable(actualType: DataType, expectedType: MethodArgType) =
    byDataType(actualType).map((t) => t == expectedType).getOrElse(false)
```

jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator$Test.scala
```diff
@@ -0,0 +1,204 @@
+package io.kaitai.struct.translators
+
+import io.kaitai.struct.datatype.DataType._
+import io.kaitai.struct.exprlang.Expressions
+import io.kaitai.struct.exprlang.Expressions.ParseException
+import io.kaitai.struct.precompile.{MethodNotFoundErrorWithArg, TypeMismatchError, WrongMethodCall}
+import io.kaitai.struct.translators.TestTypeProviders._
```
generalmimon, line 7: I'm just saving myself a bit of headache when I was unable to grasp for a moment where the `Always` thing in:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/c9dd0aa5414b2f8c1a383f8c75948f40501ff01e/jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator%24Test.scala#L12

... comes from :)

```suggestion
import io.kaitai.struct.translators.TestTypeProviders.Always
```

jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator$Test.scala
```diff
@@ -0,0 +1,204 @@
...
+      }
+
+      it("123.456e12ff") {
+        val thrown = the [ParseException] thrownBy Expressions.parse("123.456e12ff")
+        thrown.getMessage should be("Position 1:11, found \"ff\"")
+      }
+
+      it("\"foo") {
+        val thrown = the [ParseException] thrownBy Expressions.parse("\"foo")
+        thrown.getMessage should be("Position 1:5, found \"\"")
```
generalmimon, line 46: Shouldn't these invalid expressions belong in `ExpressionsSpec` ([**ExpressionsSpec.scala**](https://github.com/kaitai-io/kaitai_struct_compiler/blob/d56a6ca9a97369f458fa089ee806e12f87d38c6d/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala)) instead?

generalmimon (approved):
@GreyCat Thanks for this improvement! The code looks good to me, but I haven't tested it - hopefully it works as expected 🤞

GreyCat:
> To begin with, I just have to make one technical remark: please resist the urge to write "nameless" commit messages like the one in https://github.com/kaitai-io/kaitai_struct_compiler/commit/580e4bc0d2ec229cfdceb9c19ba69a733fb6467f.

Hey, really appreciate you calling this out. Much to my shame, I was not really aware of this integration capabilities, nor I have ever pain attention to this. Will fix.

shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
```diff
@@ -4,9 +4,137 @@ import io.kaitai.struct.datatype.DataType
...
+
+  sealed trait MethodSig {
+    def name: String
+    def expectedArgs: String
+    def accepts(argsValues: Seq[Ast.expr]): Boolean
+  }
+
+  case class MethodSig0(
+    name: String,
+    returnType: DataType,
```
generalmimon, line 57: I've just noticed that `returnType` is not used anywhere and the [`detectAttributeType`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/d56a6ca9a97369f458fa089ee806e12f87d38c6d/shared/src/main/scala/io/kaitai/struct/translators/TypeDetector.scala#L152) method still contains match cases like `case "first" | "last" | "min" | "max" => Int1Type(false)`, I suppose the `detectAttributeType` method is planned to be migrated to use these new method signatures?

GreyCat: Great observation! Yeah, that is the idea. Or at least that was the idea.

Now, though, the more I think about it, the more I understand how our story with all these checks/modularization went crazy and it's hard to untangle it:

* TypeProvider is keeping state of current parsing point for resolving fields out of "this" context, providing point in the tree, and keeping KS io.kaitai.struct.format constructs separate from io.kaitai.struct.exprlang. We can argue how bad it is that it's stateful, but for now I believe this is least of the concerns.
* TypeDetector existing by itself (and used in quite a few places, *including* CommonMethods) — but ideally this should be tied closely to a declarative method description like the one in CommonMethods, otherwise we have a lot of duplication and error stemming from that.
* `CommonMethods[T]` and other generic components that together make up a "translator" — but they for sure depend on TypeDetector (or do they?)
* `ExpressionValidator` which follows "translator" pattern in spirit (by providing `def validate(v: Ast.expr): Unit` which throws exceptions), but not in letter (i.e. it doesn't inherit from AbstractTranslator or BaseTranslator, because their basic requirement is having `def translate(v: Ast.expr): String`).
  * It still inherits `TypeDetector`.
  * It still inherits `CommonMethods[Unit]`.
  * In the hindsight, I see the idea of returning Unit and throwing exceptions as particularly bad — it should have been refactored as returning list of `CompilationProblem` or something like that, so we can accumulate those.
* `TypeValidator` (which is a precompile step) traverses everything and then selectively invoked `ExpressionValidator` on expressions to both use it as `TypeDetector` and do `validate(...)`. This one also adds layer of exception catching wrappers (which are, in all fairness, done quite poorly) and do that translation from exceptions to CompilationProblem.

So, a lot of things to think of...

jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator$Test.scala
GreyCat: Yeah, I think you're right. I was thinking of building some kind of comprehensive suite, but then I guess the point really is testing ExpressionValidator, but Expressions.parse. Let me remove these.

jvm/src/test/scala/io/kaitai/struct/translators/ExpressionValidator$Test.scala
GreyCat: Fixed, thanks!

PR #270, opened 2024-03-06
Fix occasional failed imports due to race conditions

generalmimon:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/951

Until now, KSC had occasional problems with resolving imported top-level types, reporting `unable to find type '<imported>'` errors from time to time (see https://github.com/kaitai-io/kaitai_struct/issues/951 for more details and reproduction instructions). It turned out that the importing code ran on multiple threads that were concurrently modifying/reading shared non-thread-safe `HashMap`s without any synchronization, which resulted in race conditions.

I thought it would be best to eliminate the race conditions by switching to some thread-safe counterpart to `mutable.HashMap`. After some googling, it turned out that there are two viable options, [`scala.collection.concurrent.TrieMap`](https://www.scala-lang.org/api/2.13.13/scala/collection/concurrent/TrieMap.html) and [`java.util.concurrent.ConcurrentHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html).

However, `TrieMap` doesn't seem to be implemented for Scala.js (https://github.com/scala-js/scala-js/issues/4866) and it is `final`, so we cannot use it as a base class of `ClassSpecs` instead of `mutable.HashMap` that we're using now. `ConcurrentHashMap` (despite being in the `java.util.concurrent.*` package, suggesting that it would only be available on JVM) surprisingly appears to be available in Scala.js (https://github.com/scala-js/scala-js/issues/1487), so in theory we could use it even in `shared/src/...` code, but I haven't figured out how to make `ClassSpecs` extend from it. It must be added that since Scala 2.13, the inheritance from `HashMap` was deprecated, so we'll have to find another way, but for now it works.

The unsynchronized hash map accesses in the `JavaClassSpecs.cached()` are in JVM-specific code (`jvm/src/...` folder), so I've just changed the private `relFiles` and `absFiles` fields to use `ConcurrentHashMap`, thus resolving the race conditions here. (`TrieMap` would work too - there's probably no practical difference for our use case.)

For the `LoadImports.loadImport()` method, I've just wrapped the accesses to shared `ClassSpecs` in a manual `synchronized` block. According to some on the internet, using `synchronized` is kind of discouraged due to being error-prone in favor of using existing thread-safe types or immutable types (Scala generally seems to prefer immutable types, but I can't imagine how they could be used in this case), but I believe it's the easiest way to solve the problem here.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/precompile/LoadImports.scala
+++ b/shared/src/main/scala/io/kaitai/struct/precompile/LoadImports.scala
@@ -76,8 +76,14 @@ class LoadImports(specs: ClassSpecs) {
           //
           // In theory, duplicate imports shouldn't be returned at all by
           // import* methods due to caching, but we won't rely on it here.
-          if (!specs.contains(specName)) {
-            specs(specName) = spec
+          val isNewSpec = specs.synchronized {
+            val isNew = !specs.contains(specName)
+            if (isNew) {
+              specs(specName) = spec
+            }
+            isNew
+          }
+          if (isNewSpec) {
             processClass(spec, ImportPath.updateWorkDir(workDir, impPath))
           } else {
             Log.importOps.warn(() => s"... we have that already, ignoring")
--- a/shared/src/main/scala/io/kaitai/struct/precompile/LoadImports.scala
+++ b/shared/src/main/scala/io/kaitai/struct/precompile/LoadImports.scala
@@ -76,6 +76,15 @@ class LoadImports(specs: ClassSpecs) {
           //
           // In theory, duplicate imports shouldn't be returned at all by
           // import* methods due to caching, but we won't rely on it here.
+          //
+          // The `synchronized` block is necessary because this code is run
+          // concurrently by multiple threads (each resolving different imports)
+          // and `specs` is a shared non-thread-safe `HashMap`. Without this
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/io/kaitai/struct/precompile/LoadImports.scala
```diff
@@ -76,8 +76,14 @@ class LoadImports(specs: ClassSpecs) {
...
-          if (!specs.contains(specName)) {
-            specs(specName) = spec
+          val isNewSpec = specs.synchronized {
+            val isNew = !specs.contains(specName)
+            if (isNew) {
+              specs(specName) = spec
+            }
+            isNew
+          }
+          if (isNewSpec) {
```
GreyCat, line 86: Maybe the only other thing I wanted to suggest here is potentially adding some rant in the comments here on why synchronized here is a must so nobody will ever think of removing it.

GreyCat:
Hey @generalmimon, I went ahead and merged this one, hope you're ok with that :)

generalmimon:
@GreyCat:

> Hey @generalmimon, I went ahead and merged this one, hope you're ok with that :)

Yes, thanks. I just let it sit for a while to see if you would look at it again ;)

PR #273, opened 2024-03-08
Fix `TranslatorSpec` tests

Mingun:
This PR fixes `TranslatorSpec` tests by changing expected values to the actually used.

Most commits fixes errors that was introduced during refactorings which did not update tests. This is understandable -- if the tests are already constantly falling, then few people think that they need to see if their changes break some of them.

Many tests related to that that each expression was wrapped in parenthesis even when this is not necessary. I know, the resulting code because of this can look ugly, and probably you want to remove those extra parenthesis instead of baking them into tests... but implementing such transalation is a long story and I think, it is better to fix tests now  just so you can use normal workflow where other PRs will be accepted based on passed tests instead of blindly trust to the PR author and maintainers' reviews. I think everyone will agree that this has already led to a bunch of errors that have been hidden for years. If you want to remove those extra parenthesis, then open a separate issue / PR, where the changes in the code and in the tests will be made simultaneousionally.

As a result, after this PR is only 21 failed tests (61 before) and only 9 in `TranslatorSpec` due to missing implementations of various features. I recommend disable those tests until features would be implemented, but I don't know how to do that. Anyway, this is better to do in another PR.


GreyCat:
Thanks for this big effort! I believe it's indeed better to merge this first, and then I wanted to work finally on https://github.com/kaitai-io/kaitai_struct/issues/69 — this will require a lot of reverts of these, but looks like it's worth it.

PR #274, opened 2024-03-09
Fix deprecation warnings from Scala 2.13.0

Mingun:
I just followed suggestions from compiler. This PR fixes most of warnings that reported by new compiler.

The only not addressed deprecation warnings are:
```
[warn] /home/runner/work/kaitai_struct_compiler/kaitai_struct_compiler/compiler/shared/src/main/scala/io/kaitai/struct/format/AttrSpec.scala:197:24: method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated (since 2.13.0): implicit conversions from Array to immutable.IndexedSeq are implemented by copying; use `toIndexedSeq` explicitly if you want to copy, or use the more efficient non-copying ArraySeq.unsafeWrapArray
[warn]           byteArray.map(x => Ast.expr.IntNum(x & 0xff))
[warn]                        ^
```
If I understand correctly, `.map()` returns a mutable array which is implicitly converted to immutable `Seq` by copying. I do not know what the best strategy to fix those warnings so I did not touch them.

### DIFF
--- a/js/src/main/scala/io/kaitai/struct/format/JavaScriptKSYParser.scala
+++ b/js/src/main/scala/io/kaitai/struct/format/JavaScriptKSYParser.scala
@@ -34,7 +34,7 @@ object JavaScriptKSYParser {
       case _: String | _: Int | _: Double | _: Boolean =>
         src
       case dict =>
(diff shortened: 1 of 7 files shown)


GreyCat (approved):
Thanks for all these fixes, looks good and meaningful!

PR #275, opened 2024-03-10
Ignore configuration of scalameta.metals VSCode Extension (Scala language server)

Mingun:
Today I installed new Scala LSP plugin in VSCode (scalameta.metals) and it creates many files of its configuration. I think, it is supposed to ignore them


GreyCat:
Sure, if it makes life easier — let's do that.

PR #277, opened 2024-03-13
KS expression language: optimize parenthesis generation

GreyCat:
Big, scary looking PR, but in reality it's rather straightforward which mostly addresses https://github.com/kaitai-io/kaitai_struct/issues/69 in 3 steps:

* TranslatorSpec: adjust existing tests which were ok with too many parenthesis to have optimal output, add some more tests.
  * Many of these are inspired/already available in [tests repo](https://github.com/kaitai-io/kaitai_struct_tests/), but I think it's still good to have them contained here for unit testing.
* Implement precedence tracking:
  * AbstractTranslator/BaseTranslator `translate(...)` now has optional precedence argument which influences whether to guard result with parenthesis or not
  * `genericBinOp` and `strConcat` use and respect this, basing their decision on `OPERATOR_PRECEDENCE` map
  * Postfix method invocations (`foo.method()`) can use now `translate` on `foo` with `METHOD_PRECEDENCE` which will ensure it's properly guarded against stuff like `.` method invocation if necessary.
* Adjust all the per-language `*Translator` implementations to use precedence tracking.
  * Removed forced parenthesis everywhere they were used.
  * Invoke `translate` or similar for underlying parts in a smart, precedence-aware way.

Out of scope of this PR:

* Left-associativity optimizations (e.g. `a + b + c` now gets rendered as `(a + b) + c`, because we have no idea that `+` is left associative). I'll figure out how to fix this later. Relevant tests are "silenced" and pass now, but in future we'll revert https://github.com/kaitai-io/kaitai_struct_compiler/commit/a2b3bef4eeaea1e64f66f0887b2f25fac849df18 and address it.
* Unary operators
* Comparison operators
* More testing for low precedence operators (e.g. BitAnd, BitOr, BitXor)

As I've mentioned, it doesn't solve 100% of all the problems, but it makes things much better, so I believe it's a good intermediate step. End result is expected to be like that:

### cpp_stl/_98/gcc4.8

🟢 1 tests fixed:
- `` ExprOpsParens: {"status"=>"format_build_failed", "elapsed"=>0.0, "is_kst"=>true} -> {"status"=>"passed", "elapsed"=>34.0, "is_kst"=>true} ``

### go//1.21

➕ 1 tests added:
- `` EnumToIInvalid -> {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} ``

### java//temurin17

🟢 1 tests fixed:
- `` ExprOpsParens: {"status"=>"format_build_failed", "elapsed"=>0, "is_kst"=>true} -> {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} ``

### javascript//nodejs12

🟢 1 tests fixed:
- `` ExprOpsParens: {"status"=>"failed", "elapsed"=>0.0, "failure"=>{"file_name"=>nil, "line_num"=>nil, "message"=>nil, "trace"=>"Expected values to be strictly equal:\n\n'012345' !== 10\n\nAssertionError [ERR_ASSERTION]: Expected values to be strictly equal:\n\n'012345' !== 10\n\n    at /tests/spec/javascript/test_expr_ops_parens.js:10:10\n    at /tests/helpers/javascript/testHelper.js:16:11\n    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3)"}, "is_kst"=>true} -> {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} ``

🔴 2 tests broken:
- `` BcdUserTypeBe: {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} -> {"status"=>"failed", "elapsed"=>0.0, "failure"=>{"file_name"=>nil, "line_num"=>nil, "message"=>nil, "trace"=>"Expected values to be strictly equal:\n+ actual - expected\n\n+ 22446688\n- 12345678\n\n      + expected - actual\n\n      -22446688\n      +12345678\n      \nAssertionError [ERR_ASSERTION]: Expected values to be strictly equal:\n+ actual - expected\n\n+ 22446688\n- 12345678\n    at /tests/spec/javascript/test_bcd_user_type_be.js:7:10\n    at /tests/helpers/javascript/testHelper.js:16:11\n    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3)"}, "is_kst"=>true} ``
- `` BcdUserTypeLe: {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} -> {"status"=>"failed", "elapsed"=>0.0, "failure"=>{"file_name"=>nil, "line_num"=>nil, "message"=>nil, "trace"=>"Expected values to be strictly equal:\n+ actual - expected\n\n+ 22446688\n- 12345678\n\n      + expected - actual\n\n      -22446688\n      +12345678\n      \nAssertionError [ERR_ASSERTION]: Expected values to be strictly equal:\n+ actual - expected\n\n+ 22446688\n- 12345678\n    at /tests/spec/javascript/test_bcd_user_type_le.js:7:10\n    at /tests/helpers/javascript/testHelper.js:16:11\n    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3)"}, "is_kst"=>true} ``

### lua//5.3

(no changes)

### perl//5.38

(no changes)

### python//3.11

➕ 1 tests added:
- `` ExprFstring0 -> {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} ``

### ruby//3.3

🟢 1 tests fixed:
- `` ExprOpsParens: {"status"=>"failed", "elapsed"=>0.0, "failure"=>{"file_name"=>"./spec/ruby/expr_ops_parens_spec.rb", "line_num"=>4, "message"=>"no implicit conversion of Integer into String", "trace"=>"./compiled/ruby/expr_ops_parens.rb:75:in `+'\n./compiled/ruby/expr_ops_parens.rb:75:in `str_concat_len'\n./spec/ruby/expr_ops_parens_spec.rb:10:in `block (2 levels) in <top (required)>'"}, "is_kst"=>true} -> {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true} ``

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/translators/AbstractTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/AbstractTranslator.scala
@@ -14,7 +14,13 @@ trait AbstractTranslator {
   /**
     * Translates KS expression into an expression in some target language.
     * @param v KS expression to translate
+    * @param extPrec precedence of external context of this expression:
+    *                it it's higher than inner expression we translate, we
+    *                will generate extra parenthesis to keep inner expression
+    *                safe
     * @return expression in target language as string
     */
-  def translate(v: Ast.expr): String
+  def translate(v: Ast.expr, extPrec: Int): String
+
+  def translate(v: Ast.expr): String = translate(v, 0)
 }
--- a/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
@@ -44,9 +44,10 @@ abstract class BaseTranslator(val provider: TypeProvider)
     * to assist error reporting in KSC.
     *
     * @param v KS expression to translate
+    * @param extPrec precedence of external context of this expression
     * @return expression in target language as string
     */
-  def translate(v: Ast.expr): String = {
+  def translate(v: Ast.expr, extPrec: Int): String = {
     v match {
       case Ast.expr.IntNum(n) =>
         doIntLiteral(n)
@@ -112,9 +113,9 @@ abstract class BaseTranslator(val provider: TypeProvider)
       case Ast.expr.BinOp(left: Ast.expr, op: Ast.operator, right: Ast.expr) =>
         (detectType(left), detectType(right), op) match {
           case (_: NumericType, _: NumericType, _) =>
-            numericBinOp(left, op, right)
+            genericBinOp(left, op, right, extPrec)
           case (_: StrType, _: StrType, Ast.operator.Add) =>
-            strConcat(left, right)
+            strConcat(left, right, extPrec)
           case (ltype, rtype, _) =>
             throw new TypeMismatchError(s"can't do $ltype $op $rtype")
         }
@@ -189,7 +190,8 @@ abstract class BaseTranslator(val provider: TypeProvider)
   def doEnumById(enumTypeAbs: List[String], id: String): String
 
   // Predefined methods of various types
-  def strConcat(left: Ast.expr, right: Ast.expr): String = s"${translate(left)} + ${translate(right)}"
+  def strConcat(left: Ast.expr, right: Ast.expr, extPrec: Int) =
+    genericBinOp(left, Ast.operator.Add, right, extPrec)
   def boolToInt(value: Ast.expr): String =
     doIfExp(value, Ast.expr.IntNum(1), Ast.expr.IntNum(0))
 
@@ -205,7 +207,7 @@ abstract class BaseTranslator(val provider: TypeProvider)
   // Helper that does simple "one size fits all" attribute calling, if it is useful
   // for the language
   def anyField(value: Ast.expr, attrName: String): String =
-    s"${translate(value)}.${doName(attrName)}"
+    s"${translate(value, METHOD_PRECEDENCE)}.${doName(attrName)}"
 
   // f-strings
   def doInterpolatedStringLiteral(exprs: Seq[Ast.expr]): String =
--- a/shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
@@ -136,6 +136,18 @@ abstract trait CommonMethods[T] extends TypeDetector {
   )
 
   /**
+    * Constant for precedence to use from within method call when we need to make sure parenthesis
+    * will appear when necessary. Mostly used when a language will use postfix method call on an
+    * object, like `obj.method(args)`:
+    *
+    *  - if `obj` is simple (like variable reference or a literal - `obj.method(args)` or
+    *    `42.method(args)`, no parenthesis are required.
+    *  - if `obj` is complex (like an expression `a + b`), then this will ensure framing as
+    *    `(a + b).method(args)`.
+    */
+  val METHOD_PRECEDENCE = 99
+
+  /**
     * Translates a certain attribute call (as in `foo.bar`) into a rendition
     * of expression in certain target language.
     * @note Must be kept in sync with [[TypeDetector.detectAttributeType]]
--- a/shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala
@@ -3,8 +3,41 @@ package io.kaitai.struct.translators
 import io.kaitai.struct.exprlang.Ast
 
 trait CommonOps extends AbstractTranslator {
-  def numericBinOp(left: Ast.expr, op: Ast.operator, right: Ast.expr) = {
-    s"(${translate(left)} ${binOp(op)} ${translate(right)})"
+  /**
+   * Provides operator precedence table, used for deciding whether
+   * parenthesis guarding expression are necessary or not.
+   *
+   * This is the default table, based on Python operator precedence model.
+   * This is good enough for most C-like languages. Individual languages'
+   * translators can override it if and when necessary to alter behavior.
+   *
+   * @see https://docs.python.org/3/reference/expressions.html#operator-precedence
+   */
+  val OPERATOR_PRECEDENCE = Map[Ast.operator, Int](
+    Ast.operator.Mult -> 13,
+    Ast.operator.Div -> 13,
+    Ast.operator.Mod -> 13,
+    Ast.operator.Add -> 12,
+    Ast.operator.Sub -> 12,
+    Ast.operator.LShift -> 11,
+    Ast.operator.RShift -> 11,
+    Ast.operator.BitAnd -> 10,
+    Ast.operator.BitXor -> 9,
+    Ast.operator.BitOr -> 8,
+  )
+
+  def genericBinOp(left: Ast.expr, op: Ast.operator, right: Ast.expr, extPrec: Int): String =
+    genericBinOpStr(left, op, binOp(op), right, extPrec)
+
+  def genericBinOpStr(left: Ast.expr, op: Ast.operator, opStr: String, right: Ast.expr, extPrec: Int): String = {
+    val thisPrec = OPERATOR_PRECEDENCE(op)
+    val leftStr = translate(left, thisPrec)
+    val rightStr = translate(right, thisPrec)
+    if (thisPrec <= extPrec) {
+      s"($leftStr $opStr $rightStr)"
+    } else {
+      s"$leftStr $opStr $rightStr"
+    }
   }
 
   def binOp(op: Ast.operator): String = {
--- a/shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
+++ b/shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
@@ -68,14 +68,14 @@ class RubyTranslator(provider: TypeProvider) extends BaseTranslator(provider)
   }
 
   override def arraySubscript(container: Ast.expr, idx: Ast.expr): String =
-    s"${translate(container)}[${translate(idx)}]"
+    s"${translate(container, METHOD_PRECEDENCE)}[${translate(idx)}]"
   override def doIfExp(condition: Ast.expr, ifTrue: Ast.expr, ifFalse: Ast.expr): String =
     s"(${translate(condition)} ? ${translate(ifTrue)} : ${translate(ifFalse)})"
 
   // Predefined methods of various types
   override def strToInt(s: Ast.expr, base: Ast.expr): String = {
     val baseStr = translate(base)
-    translate(s) + ".to_i" + (baseStr match {
+    s"${translate(s, METHOD_PRECEDENCE)}.to_i" + (baseStr match {
       case "10" => ""
       case _ => s"($baseStr)"
     })
@@ -83,9 +83,9 @@ class RubyTranslator(provider: TypeProvider) extends BaseTranslator(provider)
   override def enumToInt(v: Ast.expr, et: EnumType): String =
     s"${enumInverseMap(et)}[${translate(v)}]"
   override def floatToInt(v: Ast.expr): String =
-    s"(${translate(v)}).to_i"
+    s"${translate(v, METHOD_PRECEDENCE)}.to_i"
   override def intToStr(i: Ast.expr): String =
-    translate(i) + s".to_s"
+    s"${translate(i, METHOD_PRECEDENCE)}.to_s"
 
   override def bytesToStr(bytesExpr: String, encoding: String): String = {
     // We can skip "encode to UTF8" if we're 100% sure that the string we're handling is already
@@ -98,7 +98,7 @@ class RubyTranslator(provider: TypeProvider) extends BaseTranslator(provider)
   }
 
   override def bytesLength(b: Ast.expr): String =
-    s"${translate(b)}.size"
+    s"${translate(b, METHOD_PRECEDENCE)}.size"
   /**
     * Alternatives considered:
     *
@@ -106,34 +106,34 @@ class RubyTranslator(provider: TypeProvider) extends BaseTranslator(provider)
     * * value.bytes[0] => 8303
     */
   override def bytesSubscript(container: Ast.expr, idx: Ast.expr): String =
-    s"${translate(container)}[${translate(idx)}].ord"
+    s"${translate(container, METHOD_PRECEDENCE)}[${translate(idx)}].ord"
   override def bytesFirst(b: Ast.expr): String =
-    s"${translate(b)}[0].ord"
+    s"${translate(b, METHOD_PRECEDENCE)}[0].ord"
   override def bytesLast(b: Ast.expr): String =
-    s"${translate(b)}[-1].ord"
+    s"${translate(b, METHOD_PRECEDENCE)}[-1].ord"
   override def bytesMin(b: Ast.expr): String =
-    s"${translate(b)}.bytes.min"
+    s"${translate(b, METHOD_PRECEDENCE)}.bytes.min"
   override def bytesMax(b: Ast.expr): String =
-    s"${translate(b)}.bytes.max"
+    s"${translate(b, METHOD_PRECEDENCE)}.bytes.max"
 
   override def strLength(s: Ast.expr): String =
-    s"${translate(s)}.size"
+    s"${translate(s, METHOD_PRECEDENCE)}.size"
   override def strReverse(s: Ast.expr): String =
-    s"${translate(s)}.reverse"
+    s"${translate(s, METHOD_PRECEDENCE)}.reverse"
   override def strSubstring(s: Ast.expr, from: Ast.expr, to: Ast.expr): String =
-    s"${translate(s)}[${translate(from)}..(${translate(to)} - 1)]"
+    s"${translate(s, METHOD_PRECEDENCE)}[${translate(from)}..${genericBinOp(to, Ast.operator.Sub, Ast.expr.IntNum(1), 0)}]"
 
   override def arrayFirst(a: Ast.expr): String =
-    s"${translate(a)}.first"
+    s"${translate(a, METHOD_PRECEDENCE)}.first"
   override def arrayLast(a: Ast.expr): String =
-    s"${translate(a)}.last"
+    s"${translate(a, METHOD_PRECEDENCE)}.last"
   override def arraySize(a: Ast.expr): String =
-    s"${translate(a)}.length"
+    s"${translate(a, METHOD_PRECEDENCE)}.length"
   override def arrayMin(a: Ast.expr): String =
-    s"${translate(a)}.min"
+    s"${translate(a, METHOD_PRECEDENCE)}.min"
   override def arrayMax(a: Ast.expr): String =
-    s"${translate(a)}.max"
+    s"${translate(a, METHOD_PRECEDENCE)}.max"
 
   override def kaitaiStreamEof(value: Ast.expr): String =
-    s"${translate(value)}.eof?"
+    s"${translate(value, METHOD_PRECEDENCE)}.eof?"
 }
--- a/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
@@ -15,91 +15,99 @@ import scala.collection.immutable.ListMap
 
 class TranslatorSpec extends AnyFunSpec {
   describe("integer literals + unary minus") {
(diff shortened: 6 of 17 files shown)


generalmimon:
Commit https://github.com/kaitai-io/kaitai_struct_compiler/pull/277/commits/fec28262c9d22e9a1337898b3dc26de6e62e7dc6 has a typo in the commit title (it shouldn't be <code>METHOD_PRECE~~NDECE~~</code>, but <code>METHOD_PRECE<ins>**DENCE**</ins></code>).

shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
```diff
@@ -98,42 +98,42 @@ class RubyTranslator(provider: TypeProvider) extends BaseTranslator(provider)
...
   override def arrayMin(a: Ast.expr): String =
-    s"${translate(a)}.min"
+    s"${translate(a, METHOD_PRECEDENCE)}.min"
   override def arrayMax(a: Ast.expr): String =
-    s"${translate(a)}.max"
+    s"${translate(a, METHOD_PRECEDENCE)}.max"
 
   override def kaitaiStreamEof(value: Ast.expr): String =
-    s"${translate(value)}.eof?"
+    s"${translate(value, METHOD_PRECEDENCE)}.eof?"
```
generalmimon, line 138: This type of change in `kaitaiStreamEof` is done in `RubyTranslator` and `JavaScriptTranslator` but not in `GoTranslator`, `LuaTranslator` or `PythonTranslator` - this is inconsistent at least.

I don't think it makes an observable difference in the current state, because the only kind of expression yielding a stream I can think of that requires wrapping into parentheses is a ternary expression (like `(cond ? a._io : b._io).eof`), which is currently always generated in parentheses `(a ? b : c)` and is not affected by external precedence.

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
```diff
@@ -44,9 +44,10 @@ abstract class BaseTranslator(val provider: TypeProvider)
     * to assist error reporting in KSC.
     *
     * @param v KS expression to translate
+    * @param extPrec precedence of external context of this expression
     * @return expression in target language as string
     */
-  def translate(v: Ast.expr): String = {
+  def translate(v: Ast.expr, extPrec: Int = 0): String = {
```
generalmimon, line 50: Actually, in any case, I'm not sure why you're specifying the default value for `extPrec` here as well, isn't it enough to specify it just in `AbstractTranslator`?

shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
```diff
@@ -98,42 +98,42 @@ class RubyTranslator(provider: TypeProvider) extends BaseTranslator(provider)
   }
 
   override def bytesLength(b: Ast.expr): String =
-    s"${translate(b)}.size"
+    s"${translate(b, METHOD_PRECEDENCE)}.size"
```
generalmimon, line 101: Once again, this change in `bytesLength` is done in `RubyTranslator` and `JavaTranslator`, but not in `CppTranslator` or `CSharpTranslator` 🤷‍♂️

I'd rather just see a consistent set of methods that are updated everywhere.

shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
```diff
@@ -135,6 +135,18 @@ abstract trait CommonMethods[T] extends TypeDetector {
...
+    * Constant for precedence to use from within method call when we need to make sure parenthesis
+    * will appear when necessary. Mostly used when a language will use postfix method call on an
+    * object, like `obj.method(args)`:
+    *
+    *  - if `obj` is simple (like variable reference or a literal - `obj.method(args)` or
+    *    `42.method(args)`, no parenthesis are required.
+    *  - if `obj` is complex (like an expression `a + b`), then this will ensure framing as
+    *    `(a + b).method(args)`.
+    */
+  val METHOD_PRECEDENCE = 99
```
generalmimon, line 148: I think it would rather make sense to have all precedence constants in a single file rather than have them scattered, given that they're compared with each other and are thus expected to follow a certain order across these different files etc.

Also, I'm thinking it would be probably better to use something more type-safe than plain integers (`case class`/`case object`/`sealed trait` or how these things are called in Scala, it all sounds similar to me ;) ). That would actually also eliminate the urge to use the magic `0` without any obvious meaning everywhere (I always have to think for a while, does it mean lowest priority or highest priority, or what purpose does it serve?) and replace that with a proper `case object` with an appropriate name (e.g. `NoPrecedence` or something, I don't know).

shared/src/main/scala/io/kaitai/struct/translators/AbstractTranslator.scala
```diff
@@ -14,7 +14,13 @@ trait AbstractTranslator {
...
+    * @param extPrec precedence of external context of this expression:
+    *                it it's higher than inner expression we translate, we
+    *                will generate extra parenthesis to keep inner expression
+    *                safe
     * @return expression in target language as string
     */
-  def translate(v: Ast.expr): String
+  def translate(v: Ast.expr, extPrec: Int): String
+
+  def translate(v: Ast.expr): String = translate(v, 0)
```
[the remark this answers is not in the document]
generalmimon: Actually, the reason the `BcdUserTypeBe` test broke in JavaScript as you mention in your [initial comment](https://github.com/kaitai-io/kaitai_struct_compiler/pull/277#issue-2184766170) exactly demonstrates how this unsafe default of `extPrec = 0` can be a problem:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/a2b3bef4eeaea1e64f66f0887b2f25fac849df18/shared/src/main/scala/io/kaitai/struct/translators/JavaScriptTranslator.scala#L33-L34

This causes the following:

```ksy
meta:
  id: shl_and
instances:
  v:
    value: '(0b01 & 0b10) >> 1'
```

... to be translated as:

```js
      this._m_v = (1 & 2 >>> 1);
```

... which is interpreted as `1 & (2 >>> 1)` by JavaScript and gives an incorrect result of `1` (while the expected result is `0`, since `0b01 & 0b10` is obviously zero and that doesn't change by shifting it).

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
```diff
@@ -189,7 +190,8 @@ abstract class BaseTranslator(val provider: TypeProvider)
   def doEnumById(enumTypeAbs: List[String], id: String): String
 
   // Predefined methods of various types
-  def strConcat(left: Ast.expr, right: Ast.expr): String = s"${translate(left)} + ${translate(right)}"
+  def strConcat(left: Ast.expr, right: Ast.expr, extPrec: Int) =
+    numericBinOp(left, Ast.operator.Add, right, extPrec)
```
Mingun, line 194: It seems quite confusing that operation on string called _numeric_ binary operation.

shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala
```diff
@@ -3,8 +3,37 @@ package io.kaitai.struct.translators
...
+    Ast.operator.Div -> 13,
+    Ast.operator.Mod -> 13,
+    Ast.operator.Add -> 12,
+    Ast.operator.Sub -> 12,
+    Ast.operator.LShift -> 11,
+    Ast.operator.RShift -> 11,
+    Ast.operator.BitAnd -> 10,
+    Ast.operator.BitXor -> 9,
+    Ast.operator.BitOr -> 8,
+  )
```
Mingun, line 23: I'm not sure that we should use the same precedence map for every language. If I understand correctly, this is precedence map of target language (which could vary for different languages) and not a precedence map of Kaitai Struct expressing language.

If I'm wrong and this is a precedence map of Kaitai Struct expressing language, then it could be better to define precedence in definition of corresponding case classes

shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
```diff
@@ -135,6 +135,18 @@ abstract trait CommonMethods[T] extends TypeDetector {
     ),
   )
 
+  /**
+    * Constant for precedence to use from within method call when we need to make sure parenthesis
+    * will appear when necessary. Mostly used when a language will use postfix method call on an
+    * object, like `obj.method(args)`:
+    *
+    *  - if `obj` is simple (like variable reference or a literal - `obj.method(args)` or
+    *    `42.method(args)`, no parenthesis are required.
```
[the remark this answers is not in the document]
GreyCat: My perspective on this is basically much simpler: we won't be able to theoretically map all these quirks of all possible languages. Let's have this covered by the tests, and, when practical, let's try to build up some common groups.

JavaScript is clearly an outlier here, so my logic is "let's just wrap it in invocation time for now, and forget about that". If and when we'll get to it — yeah, in theory we can expand "precedence" concept to something more complicated.

shared/src/main/scala/io/kaitai/struct/translators/AbstractTranslator.scala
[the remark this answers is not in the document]
GreyCat: Good point and thanks for finding the root cause for BCD-related tests. This case is definitely something we'd want to get in for integer operation tests in TranslatorSpec too and fix that.

> Does translate need to have a default value for extPrec at all?

Good question. When I've started experimenting with it, not having default value proven to be requiring *massive* changes that will be hard to complete and review. In general, you're right, all these places need to be reviewed and we should (1) cover them with tests, (2) consciously decide whether these need braces or not.

Let me see and experiment what harm/good will do to make the default super high (METHOD_PRECEDENCE).

shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
GreyCat: Good point and a good idea for further testing. Let me do the same change everywhere and add the test in future.

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
GreyCat: You're right, good catch. That's a leftover of previous implementation I forgot to remove. Fixed.

shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
GreyCat: Good point, let me fix that at least and add some tests. I was under impression that this will be only affected when base type has `+` operator (strings do have it, byte arrays don't), but you gave a perfect idea that it will be also affecting ternary operator, so let me start building that.

shared/src/main/scala/io/kaitai/struct/translators/CommonMethods.scala
GreyCat: Let me think if this is achievable. The problem here is that these are not set in stone, they are supposed to be (slightly) varying across the languages. Therefore, by that definition, they will be represented in more than 1 file anyway: in some common file + per-language file.

Wrapping them into some special type indeed will make things more "official" and likely typesafe, but I'm not sure if we'll get a lot of value from it: in the end of the day, this is all about comparisons, e.g. something like:

```scala
trait Precedence extends Ordering[Precedence] {
  def compare(a: Precedence, b: Precedence) = ???
}
case object LowPrecedence extends Precedence
case object AddSubPrecedence extends Precedence
case object MulDivPrecedence extends Precedence
```

That `compare(a, b)` can be implemented either as:

* Lots of match/case statements => looks super ugly to me and likely very error prone
* Map all these to integers and compare integers

"Mapping to integers" will basically just overcomplicate what we're already doing: instead of mapping "Ast.expr.operator -> Int", we'll do "Ast.expr.operator -> Precedence -> Int".

For sake of mentioning it, there is one other solution based on auto-enumeration constant generation:

```
object Precedence extends Enumeration {
  val LowPrecedence, AddSubPrecedence, MulDivPrecedence, MethodPrecedence = Value

  implicit val ordering: Ordering[Value] = new Ordering[Value] {
    def compare(x: Value, y: Value): Int = {
      x.id - y.id
    }
  }
}
```

However, all of these will anyhow complicate overriding these on per-language basis.

I'd rather have flexibility using simple numbers (and this probably calls to multiply proposed integer values by x10 and leave gaps between them), rather than deal with repeating/modifying this verbose construct on per-language basis.

shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala
GreyCat: It is in BaseTranslator, and it's easy to override if need be for specific language. As a docstring says, the default is "good enough for most C-like languages".

So far I haven't found any single instance where this will be needed, but I assume at some point we'll encounter something like ternary operator vs bitshifts vs bit operators vs comparison could be treated very differently.

I recall at least that Pascal requires this:

```
(a < b) and (b < c)
```

while C-like languages normally have this:

```
a < b && b < c
```

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
GreyCat: I kinda agree... Do you think something like

```scala
def numericBinOp(left, right, extPrec) = genericBinOp(left, right, extPrec)
def strConcat(left, right, extPrec) = genericBinOp(left, right, extPrec)
```

... will be better?

generalmimon:
> I can prepare this right now (force-pushing), or do it later when merging — what do you prefer?

I don't care - whatever is more convenient for you 🤷‍♂️

It's true that there are currently arguably too many commits, so it's easier to review the whole thing than go through each individual commit. So squashing them into several larger units makes sense to me.

shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala
Mingun: Ok, I just have an impression, that `val` is a field that cannot be overridden, but I'm not a Scala expert. I think, however, that a comment, explaining that this is a precedence in a target language and it can be overridden in specific languages will be good addition.

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
Mingun: Well, just a comment explaining why we call a member named `numeric` for `str` operation would be enough. For example, something like "logic the same, we are just reuse the best fit method". Your variant is also good.

shared/src/main/scala/io/kaitai/struct/translators/CommonOps.scala
GreyCat: To be honest, I don't recall as well if `val` can be overridden or not, but the point is we can do it in future if and when necessary. Noted about the comment, let me add that.

shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
GreyCat: Fixed in 96a26f0e7084d1f4caf86d816279f2692c469a26

generalmimon:
While testing this, I found a few existing bugs in some language-specific translator method impls, but they're most likely out of scope of this PR (they are certainly not new):

1. missing parentheses in `LuaTranslator.floatToInt`

   https://github.com/kaitai-io/kaitai_struct_compiler/blob/96a26f0e7084d1f4caf86d816279f2692c469a26/shared/src/main/scala/io/kaitai/struct/translators/LuaTranslator.scala#L104-L105
   
   ```ksy
   instances:
     l11_r0_lsig0_rsig0:
       value: 10.6.to_i + 15
   ```
   
   ```lua
     self._m_l11_r0_lsig0_rsig0 = (10.6 > 0) and math.floor(10.6) or math.ceil(10.6) + 15
   ```
   
   ```
   spec/lua/test_expr_gen.lua:287: expected: 25, actual: 10
   ```

2. missing parentheses in `PerlTranslator.strToInt` for base 10

   https://github.com/kaitai-io/kaitai_struct_compiler/blob/96a26f0e7084d1f4caf86d816279f2692c469a26/shared/src/main/scala/io/kaitai/struct/translators/PerlTranslator.scala#L124-L125
   
   ```ksy
   instances:
     l11_r2_lsig1_rsig0:
       value: '"988".to_i * 2'
   ```
   
   ```perl
       $self->{l11_r2_lsig1_rsig0} = "988" + 0 * 2;
   ```
   
   ```
   #   Failed test 'Equals'
   #   at spec/perl/TestExprGen.t line 311.
   #   (in spec::perl::TestExprGen->test_expr_gen)
   #          got: '988'
   #     expected: '1976'
   ```

3. `RubyTranslator.strSubstring` implementation doesn't work for `end = 0`

   https://github.com/kaitai-io/kaitai_struct_compiler/blob/96a26f0e7084d1f4caf86d816279f2692c469a26/shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala#L123-L124
   
   ```ksy
     substr_from0to0:
       value: '"30293612".substring(0, 0)'
   ```
   
   ```ruby
       @substr_from0to0 = "30293612"[0..0 - 1]
   ```
   
   ```
   expected: ""
        got: "30293612"
   
   (compared using ==)
   ```

shared/src/main/scala/io/kaitai/struct/translators/AbstractTranslator.scala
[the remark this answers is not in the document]
generalmimon: > Let me see and experiment what harm/good will do to make the default super high (METHOD_PRECEDENCE).

Sounds good. For the record, I've discovered another concrete case where the assumed precedence of `0` poses a problem, namely here:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/6aef6faeb00d4800d493460bc0f83483e60bad01/shared/src/main/scala/io/kaitai/struct/translators/MinSignedIntegers.scala#L22-L29

https://github.com/kaitai-io/kaitai_struct_compiler/blob/96a26f0e7084d1f4caf86d816279f2692c469a26/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala#L98-L105

```ksy
meta:
  id: test
instances:
  min_b64_div2:
    value: -0x8000_0000_0000_0000 / 2
```

```console
$ kaitai-struct-compiler -t lua test.ksy
```

`test.lua`

```lua
  self._m_x = math.floor(-9223372036854775807 - 1 / 2)
```

https://tio.run/##yylN/P@/oCgzr0QjN7EkQy8tJz@/SEPX0sjI2NjcyMDYzMLUxNzc1MLAXEFXwVBBX8FIU/P/fwA

```lua
print(math.floor(-9223372036854775807 - 1 / 2)) -- => -9223372036854775808, but expected -4611686018427387904
                                                --    (a.k.a. -0x4000_0000_0000_0000)
```

Mingun:
An idea: introduce new `translateMethod(expr)` helper method to do `translate(expr, METHOD_PRECEDENCE)`. I seems that will be more readable.

shared/src/main/scala/io/kaitai/struct/translators/RubyTranslator.scala
GreyCat: Fixed the problem in 59decb30, but haven't added any tests yet, as I intentionally don't want to go into rabbit hole of ternary operators in this PR just yet.

shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala
GreyCat: Ended up renaming it into `genericBinOp` altogether, because now it really makes a whole less sense to separate this truly only for numericals.

shared/src/main/scala/io/kaitai/struct/translators/AbstractTranslator.scala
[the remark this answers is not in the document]
GreyCat: Yep, all these are supposed to go into https://github.com/kaitai-io/kaitai_struct_compiler/blob/optimal_expr_v2/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala#L110

GreyCat:
Hey folks, @generalmimon @Mingun, I went ahead, reshaped this into 3 commits and merged. It looks like it's already bringing some value, fixes tons of problems and (at least according to my test runs) it doesn't make anything worse than it was.

I will take your feedback on additional test cases in further PRs and commits — likely next targets would be (roughly in this order):

* comparison operators and boolean operators
* ternary operator
* integer operations with extreme integer constants
* unary operators

PR #278, opened 2024-03-15
Fix excess warnings about inherited encoding attributes

Mingun:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/1088.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames$Test.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames$Test.scala
@@ -1,5 +1,8 @@
 package io.kaitai.struct.precompile
 
+import io.kaitai.struct.datatype.DataType.{CalcBytesType, StrFromBytesType}
+import io.kaitai.struct.exprlang.Ast
+import io.kaitai.struct.format._
 import io.kaitai.struct.problems._
 import org.scalatest.funspec.AnyFunSpec
 import org.scalatest.matchers.should.Matchers._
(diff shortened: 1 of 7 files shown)


jvm/src/test/scala/io/kaitai/struct/precompile/CanonicalizeEncodingNames$Test.scala
```diff
@@ -44,5 +47,66 @@ class CanonicalizeEncodingNames$Test extends AnyFunSpec {
...
+
+        val noProblems = CanonicalizeEncodingNames.canonicalizeMember(ParseInstanceSpec(
+          id = InstanceIdentifier("parse_instance"),
+          path = List("parse", "instance"),
+          dataType = StrFromBytesType(CalcBytesType, "utf-8", true),
+        )).toList
+        noProblems should be(List())
+      }
+
+      it("in value instance") {
```
Mingun, line 90: Not sure if testing value instances is necessary, they cannot have `StrFromBytesType` string type, I guess. I simply testing every implementation of `MemberSpec`

shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
```diff
@@ -400,7 +413,7 @@ object DataType {
           }
 
           val bat = arg2.getByteArrayType(path)
-          StrFromBytesType(bat, enc)
+          StrFromBytesType(bat, enc, arg.encoding.isEmpty)
```
GreyCat, line 416: While this indeed solves the problem, I'm not so sure it's the right approach in the long run.

Basically, what happens now is:

* We do YAML → internal in-memory spec representation, and during that conversion we simultaneously do resolution. It would have been a perfect point to do the check as well and signal if we don't like anything about the encoding, but, sadly, we don't have "warnings" emission at that stage at all. It's either hard stop with an exception, or "all good".
* Afterwards, we do "precompilation" steps, which, in theory, are supposed to do all these warning/errors collections, and conversions as well (e.g. class names resolution or parent types mapping happens as precompile step) — but, alas, at this point the damage is already done and we've already cemented on which encoding we propagated everywhere.

Your solution basically tries to salvage this situation by propagating information "hey, this is not genuine encoding, but something coming from meta" into the type definition to be able to distinguish at the next steps.

My take on this is we should rather fix the root cause. My ideal solution would be:

* YAML → internal spec representation becomes as simple as possible.
  * If possible, we should indeed just get this down to bare minimum stuff, where the only possible outcome is something not matching the schema (=syntax error), thus completely impossible to recover, thus exception is the only way.
  * If we find ourselves being more "forgiving", maybe we should introduce "soft" problem reporting with list of CompilationProblems from that stage.
* Everything related to semantical checks or resolutions should happen *after*, cleanly as separate precompile step. If this will involve modifying the spec tree, that's alright. This step will be able to report problems in soft manner, and thus be able to easily distinguish between meta/in-attribute encoding names spec, etc.

That's obviously a much bigger change. @Mingun, @generalmimon, what do you think?

Mingun: I agree that this is the right way to do that. For example, I use exactly that approach in my [reimplementation](https://github.com/Mingun/ksc-rs) of ksc in Rust in `parser` and `model` modules (I should find a time to resurrect project!). At the same time this requires pretty big changes in the compiler, so this can take significant time when this will be implemented.

So we have choice to merge the not ideal fix today, increase legacy code base and replace it in the future with the right version, or wait when internals will be completely rewritten to make the right fix possible and all that time have this bug.

I would prefer the first variant. That way we also can increase test coverage so it should be more easy in the future rewrite the things in the right way.

generalmimon: I agree with both of you. If a non-canonical encoding (like `Ascii`) is specified only once at `/meta/encoding` and then used by 4 occurrences of `type: str` in a .ksy spec, then it would make sense to fix it only once at `/meta/encoding`, and only distribute the already fixed encoding (e.g. `ASCII`) to individual `str` types, instead of fixing it anew 5 times and suppressing the warning for `str` types that have merely inherited the (non-canonical) `/meta/encoding`.

This would mean that the propagation of `encoding` would have to be moved from the YAML parsing component to a precompile step, as @GreyCat pointed out. I think that makes total sense, and it would have other benefits as well.

For example, the `error: string type, but no encoding found` could then be reported for each `type: str` (not only for the first one), because technically it is a problem that concerns all of them:

```ksy
meta:
  id: missing_encoding
seq:
  - id: a
    type: strz
  - id: b
    type: strz
```

Currently, only the first `type: str` (in this case `/seq/0`) is mentioned in the error message (because the feature of reporting multiple problems at once is exclusive to precompilation, the YAML parsing phase stops at the first error):

```console
$ kaitai-struct-compiler -t python missing_encoding.ksy
missing_encoding.ksy: /seq/0:
        error: string type, but no encoding found
```

But strictly speaking, that is incomplete, because if you just fix it on the referenced field:

```diff
 seq:
   - id: a
     type: strz
+    encoding: ASCII
   - id: b
     type: strz
```

... now you'll obviously get the same error on `/seq/1`.

---

But anyway, I like the simplicity of the @Mingun's fix in the current codebase, and the fact that it's simple also means it'll be easy to refactor to implement it in a more clean way later on.

Probably just a simple `// FIXME: ...` comment in the code next to `isEncodingDerived` that states that we'd like to eventually get rid of it (as soon as the architecture allows it) would be appropriate, though, so we don't forget.

Mingun: I've added the FIXME comment to definition of new parameter with reference to this discussion. Actually, we could consider current `DataType` as intermediate representation and we should introduce new types for final representation. This also would allow us to get rid of warnings
```
[warn] /home/runner/work/kaitai_struct_compiler/kaitai_struct_compiler/compiler/shared/src/main/scala/io/kaitai/struct/languages/CSharpCompiler.scala:358:5: Exhaustivity analysis reached max recursion depth, not all missing cases are reported.
[warn] (Please try with scalac -Ypatmat-exhaust-depth 40 or -Ypatmat-exhaust-depth off.)
[warn]     dataType match {
[warn]     ^
```
because part of types are handled outside of `parseExpr` methods (where such warnings are generated). I've made investigation and possible types here is any `DataType` except `SwitchType`, `StrFromBytesType`, and `EnumType`.

Mingun:
The 4 new failures because of recent changes in kaitai_struct_tests [which are expected right now](https://github.com/kaitai-io/kaitai_struct_tests/commit/9d518e6816f2a18d6459423aad23c77450cb6ab6).

shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
GreyCat: Perfect! Thanks for the discussion, folks, I'm confident we can merge then.

GreyCat (approved):

PR #280, opened 2024-03-17
Calculate parent for types that does not used in the current spec

Mingun:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/961

I have not decided yet, how to test that, so draft for now


generalmimon (approved):
@Mingun Thanks!

generalmimon:
> I have not decided yet, how to test that

I added a new `formats_err` test called `nav_parent_multi_one_unused` (see https://github.com/kaitai-io/kaitai_struct_tests/commit/ede1105eb313c222b7e1d9f1c357cfaac4b6a964).

Mingun:
Thanks. I think, it also need to add explicit test for `_parent: false` case as well, which should compiled successfully

PR #283, opened 2024-03-20
Pass `_root` and `_parent` to all local and no external types

generalmimon:
* Fixes https://github.com/kaitai-io/kaitai_struct/issues/1089

  Previously, `_root` and `_parent` weren't passed to the local top-level type, which is fixed by this PR.

  This was demonstrated by tests NavRootRecursive and NavParentRecursive added in https://github.com/kaitai-io/kaitai_struct_tests/commit/5941262e

* Fixes https://github.com/kaitai-io/kaitai_struct/issues/963

  `_root` and `_parent` were incorrectly passed to nested types of imported specs (i.e. external types).

  This was demonstrated by asserts in the [KST spec of test NestedTypesImport](https://github.com/kaitai-io/kaitai_struct_tests/blob/d07deb4c/spec/ks/nested_types_import.kst#L17-L32).

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
@@ -164,10 +164,12 @@ object DataType {
   ) extends StructType {
     var classSpec: Option[ClassSpec] = None
     /**
-     * Determines whether the user type represented by this `UserType` instance
-     * is external from the perspective of the given `ClassSpec` in which it is
-     * used (via `seq`, `instances` or `params`).
-     */
+      * Determines whether the user type represented by this `UserType` instance
+      * is external from the perspective of the given `ClassSpec` in which it is
+      * used (via `seq`, `instances` or `params`).
+      * @param curClass class spec from which the local/external relationship
+      * should be evaluated
+      */
     def isExternal(curClass: ClassSpec): Boolean =
       classSpec.get.isExternal(curClass)
     def isOpaque = {
--- a/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
@@ -163,6 +163,13 @@ object DataType {
     var args: Seq[Ast.expr]
   ) extends StructType {
     var classSpec: Option[ClassSpec] = None
+    /**
+     * Determines whether the user type represented by this `UserType` instance
+     * is external from the perspective of the given `ClassSpec` in which it is
+     * used (via `seq`, `instances` or `params`).
+     */
+    def isExternal(curClass: ClassSpec): Boolean =
+      classSpec.get.isExternal(curClass)
     def isOpaque = {
       val cs = classSpec.get
       cs.isTopLevel || cs.meta.isOpaque
--- a/shared/src/main/scala/io/kaitai/struct/precompile/ParentTypes.scala
+++ b/shared/src/main/scala/io/kaitai/struct/precompile/ParentTypes.scala
@@ -68,7 +68,6 @@ class ParentTypes(classSpecs: ClassSpecs) {
   }
 
   def markupParentAs(curClass: ClassSpec, ut: UserType): Unit = {
-    Log.typeProcParent.info(() => s"..... class=$ut has parent=${curClass.nameAsStr}")
     ut.classSpec match {
       case Some(usedClass) =>
         markupParentAs(curClass, usedClass)
@@ -79,6 +78,10 @@ class ParentTypes(classSpecs: ClassSpecs) {
   }
 
   def markupParentAs(parent: ClassSpec, child: ClassSpec): Unit = {
+    // Don't allow type usages across spec boundaries to affect parent resolution
+    if (child.isExternal(parent))
+      return
+    Log.typeProcParent.info(() => s"..... class=${child.nameAsStr} has parent=${parent.nameAsStr}")
     child.parentClass match {
       case UnknownClassSpec =>
         child.parentClass = parent
--- a/shared/src/main/scala/io/kaitai/struct/precompile/ParentTypes.scala
+++ b/shared/src/main/scala/io/kaitai/struct/precompile/ParentTypes.scala
@@ -79,8 +79,10 @@ class ParentTypes(classSpecs: ClassSpecs) {
 
   def markupParentAs(parent: ClassSpec, child: ClassSpec): Unit = {
     // Don't allow type usages across spec boundaries to affect parent resolution
-    if (child.isExternal(parent))
+    if (child.isExternal(parent)) {
+      Log.typeProcParent.info(() => s"..... cross-spec usage of class=${child.nameAsStr} from parent=${parent.nameAsStr} ignored")
       return
+    }
     Log.typeProcParent.info(() => s"..... class=${child.nameAsStr} has parent=${parent.nameAsStr}")
     child.parentClass match {
       case UnknownClassSpec =>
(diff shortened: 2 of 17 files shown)


Mingun:
I correctly understand, that the 3rd commit disables passing `_parent` to the types defined in imported specs? If yes, why? It is clear for `_root` because you just does not have the correct type for it in scope, but `_parent` still can be passed as `this`, right?

Using your example from the first commit:
```ksy
# main.ksy
meta:
  id: main
  imports:
    - imported
seq:
  - id: foo
    type: imported::nested_type

# imported.ksy
meta:
  id: imported
types:
  nested_type: {}
```

I think, that these changes should be expected for Java target because of your fix (you mentioned that Java does not pass `_parent` today):

```diff
 // Imported.java
 public class Imported extends KaitaiStruct {
     public static class NestedType extends KaitaiStruct {
-        public NestedType(KaitaiStream _io, Main _parent, Imported _root) { /* ... */ }
+        public NestedType(KaitaiStream _io, KaitaiStruct _parent, Imported _root) { /* ... */ }
     }
 }
 // Main.java
 public class Main extends KaitaiStruct {
     void _read() {
-        this.foo = new Imported.NestedType(this.io);
+        this.foo = new Imported.NestedType(this.io, this, null);
     }
 }
```

generalmimon:
@Mingun:

> I correctly understand, that the 3rd commit disables passing `_parent` to the types defined in imported specs?

Yes.

> It is clear for `_root` because you just does not have the correct type for it in scope, but `_parent` still can be passed as `this`, right?

Well, not in general. The change in https://github.com/kaitai-io/kaitai_struct_compiler/pull/283/commits/10b05885b3c5e9f69bbf2decaaa10aca9a632a8f means that cross-spec type usages don't influence the `_parent` type, so you cannot assume that `this` in the importing type has a type compatible with the target `_parent` parameter. In particular, passing `_parent` between different .ksy specs would be a problem e.g. in this case:

`main.ksy`

```ksy
meta:
  id: main
  imports:
    - imported
seq:
  - id: foo
    type: imported::nested_type(false)
```

`imported.ksy`

```ksy
meta:
  id: imported
seq:
  - id: len_body
    type: u1
  - id: wrapper
    type: nested_type(true)
types:
  nested_type:
    params:
      - id: use_parent
        type: bool
    seq:
      - id: body
        size: _parent.len_body
        if: use_parent
```

(The `use_parent` parameter used in `if` isn't really necessary for the demo, it's just to handwave the somewhat questionable runtime behavior that otherwise the instantiation of the `main` module would always result in a runtime error.)

As you can see, `imported::nested_type` relies on having `imported` as `_parent`, so if you try to pass `main` as `_parent` instead, that will be a compile error in statically typed languages, as you can imagine:

```console
$ javac -classpath /c/temp/kaitai_struct/runtime/java/target/kaitai-struct-runtime-0.11-SNAPSHOT.jar Main.java Imported.java
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
Main.java:28: error: incompatible types: Main cannot be converted to Imported
        this.foo = new Imported.NestedType(this._io, this, false);
                                                     ^
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
```

... since the possible constructor signatures of class `Imported.NestedType` are the following:

```java
    public static class NestedType extends KaitaiStruct {

        public NestedType(KaitaiStream _io, boolean useParent) {
            this(_io, null, null, useParent);
        }

        public NestedType(KaitaiStream _io, Imported _parent, boolean useParent) {
            this(_io, _parent, null, useParent);
        }

        public NestedType(KaitaiStream _io, Imported _parent, Imported _root, boolean useParent) {
            // ...
```

---

In conclusion, I'm convinced we need https://github.com/kaitai-io/kaitai_struct_compiler/pull/283/commits/10b05885b3c5e9f69bbf2decaaa10aca9a632a8f because letting different .ksy specs affect the derived `_parent` types of each other seems to be a big can of worms (it means that a .ksy spec behaves differently depending on whether it's imported it from somewhere or compiled directly, namely a .ksy spec could compile without errors directly and with errors when imported from somewhere or vice versa, or the generated code of the imported module would depend on where the module is imported from etc.), and not actually passing `_parent` between specs is a direct logical consequence of that.

So, as @GreyCat pointed out in https://github.com/kaitai-io/kaitai_struct/issues/71#issuecomment-366965959, if someone feels like they need to pass `_parent` between different .ksy specs, it's better to pass what they need using `params`.

Keeping both `_root` and `_parent` local just seems to be the only sane way to me once you think about the consequences of doing otherwise. Also, this was already discussed, e.g.:

https://github.com/kaitai-io/kaitai_struct/issues/71#issuecomment-366955910

> If one externalizes a KSY and imports it later in some other type, how are `_root` and `_parent` supposed to work in the extern type?

https://github.com/kaitai-io/kaitai_struct/issues/71#issuecomment-366963986

> Modules are modular. If they accessed parent types, for example using _root, they would be not modular, so they would not be modules. And this was discussed somewhere.

---

https://github.com/kaitai-io/kaitai_struct/issues/275#issuecomment-368582285

> With your example CTOR invocation, `_parent` is crossing KSY-borders and [didn't you say you don't want that](https://github.com/kaitai-io/kaitai_struct/issues/71#issuecomment-366963986) or is that only meant for `_root`? Might not be a good idea to handle both differently.

https://github.com/kaitai-io/kaitai_struct/issues/275#issuecomment-368605026

> > With your example CTOR invocation, `_parent` is crossing KSY-borders
> 
> Yeah, and that things actually bothers me as well. (...)

So I believe the consensus is clear.

In the long run, I'm not completely opposed to https://github.com/kaitai-io/kaitai_struct/issues/770, i.e. getting rid of `_parent` and `_root` completely in favor of explicit `params`. Especially the implicitly derived type of `_parent` seems to create many problems that would not exist without it. Also, you can never be sure whether both `_parent` and `_root` are actually non-`null` until runtime, etc.

Mingun:
Thanks for that so detailed explanation. Yes, falling asleep, I realized that when `_parent` is deduced to some local type of imported spec it would be impossible to pass types from the main spec there. So you implemented approach is correct. Maybe document those observations somewhere in the code?

Mingun (approved):

shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
```diff
@@ -163,6 +163,13 @@ object DataType {
     var args: Seq[Ast.expr]
   ) extends StructType {
     var classSpec: Option[ClassSpec] = None
+    /**
+     * Determines whether the user type represented by this `UserType` instance
+     * is external from the perspective of the given `ClassSpec` in which it is
+     * used (via `seq`, `instances` or `params`).
+     */
```
GreyCat, line 170: Nit: I believe up until then the rest of our codebase uses docstrings formatted this way:

```suggestion
    /**
      * Determines whether the user type represented by this `UserType` instance
      * is external from the perspective of the given `ClassSpec` in which it is
      * used (via `seq`, `instances` or `params`).
      */
```

GreyCat: Also ideally we should have `@param curClass` description here.

shared/src/main/scala/io/kaitai/struct/precompile/ParentTypes.scala
```diff
@@ -79,6 +78,10 @@ class ParentTypes(classSpecs: ClassSpecs) {
   }
 
   def markupParentAs(parent: ClassSpec, child: ClassSpec): Unit = {
+    // Don't allow type usages across spec boundaries to affect parent resolution
+    if (child.isExternal(parent))
+      return
```
GreyCat, line 83: Nit: this should ideally log something for debugging/tracing purposes so it will be easier to understand what's going on.

GreyCat (approved):

GreyCat:
Looks great, thanks for making this reality, @generalmimon!

shared/src/main/scala/io/kaitai/struct/precompile/ParentTypes.scala
generalmimon: Done.

PR #287, opened 2024-03-25
Multiply also `METHOD_PRECEDENCE` constant to match new precedence of operators

Mingun:
The last commit multiply precedence of all operators by 10x, but the constant `METHOD_PRECEDENCE` remains unchanged and as a results many tests failed.

Please, be careful and checks the CI results. I see good efforts in the last time to make CI green and I hope we can do that, but until that we need to be focuses and check results manually.

The ever-falling CI relaxes, so it's just overlooked, and then gets stupid bugs in the code.


Mingun:
Fixed in https://github.com/kaitai-io/kaitai_struct_compiler/commit/4fdf4c00d9be7a9a0bd1ddb2e6f23e201a5a097e

GreyCat:
My apologies for this mistake, thanks for noticing and producing the fix!

PR #288, opened 2024-03-25
Java: generate enums as interface + enum + class for unknown values

### LINKED ISSUE #583: Improve support for debugging by throwing on missing ENUMs.
# The problem.

At least in Java, mapping decimal values of some stream to some ENUM is done by manually creating a hash map of decimals to ENUMs and accessing that map, like in the following example:

        private static final Map<Long, ValueType> byId = new HashMap<Long, ValueType>(41);
        static {
            for (ValueType e : ValueType.values())
                byId.put(e.id(), e);
        }
        public static ValueType byId(long id) { return byId.get(id); }

What happens if the given decimal can't be found? `null` is returned, leading to NullPointerException in code accessing getters providing the ENUMs.

I have a lot of ENUMs which need to support only few values at the beginning and are extended as needed, so the exception isn't that wrong at all, because it tells me that I need to look at the parser to see if it's time to extend the ENUM once more. The problem is that the NPE doesn't tell me anything about which decimal was in the stream and stuff.

# Possible solutions.

I only see two useful solutions, either providing a fallback value for any unknown decimal or simply throwing an exception including the unknown value. Some fallback would be in line with `switch-on` which supports one as well, the exception OTOH would be more what I really need.

In my opinion, mapping ENUMs is a good candidate for validations/assertions (#435) and something like `UnexpectedDataError` in the long term. Especially because of proposed improved debugging using positional infos in exceptions (#581), throwing an exception at all in combination with positions within the stream would make recognizing the missing ENUM and deciding what to do a lot easier.

# Place mapping into runtime?

How about making ENUM-mapping itself part of the runtime? It's always the same for all of those besides that the type is different each time. This way the compiler would need to create less code and one could use all validations/assertions/exceptions available in the runtime in future.

# Any thoughts?

Mingun:
This PR implements suggestion from https://github.com/kaitai-io/kaitai_struct/issues/778 which is also enforced by [enum_invalid.ksy](https://github.com/kaitai-io/kaitai_struct_tests/blob/0c32cbfb1f600fe6c4d7fe0b65d959d732164889/formats/enum_invalid.ksy) test.

Related issues:
- Runtime changes: https://github.com/kaitai-io/kaitai_struct_java_runtime/pull/42
- Tests changes: https://github.com/kaitai-io/kaitai_struct_tests/pull/125
- Tracking issue: https://github.com/kaitai-io/kaitai_struct/issues/778
- Closes https://github.com/kaitai-io/kaitai_struct/issues/583

The new generated enum is represented as an interface with two nested classes:
- enum class `Known` with the list of known enumeration values
- class `Unknown` which represents any not-known value

The enum `enum` would be generated as:

```java
interface Enum extends IKaitaiEnum {
    public static class Unknown extends IKaitaiEnum.Unknown implements Enum {
        private Unknown(long id) { super(id); }
    }
    public enum Known implements Enum {
        Variant1(1),
        Variant2(2),
        Variant(3);

        private final long id;

        static HashMap<Long, Enum> variants = new HashMap<>(3);
        static {
            for (final Known e : Known.values()) {
                variants.put(e.id, e);
            }
        }

        private Known(long id) { this.id = id; }
        @Override
        public long id() { return this.id; }
    }

    public static Enum byId(final long id) {
        return Known.variants.computeIfAbsent(id, _id -> new Unknown(id));
    }
}
```
[Example](https://godbolt.org/z/rvsd5WEMP)

Unfortunately, it is not possible to generate enum what will implement nested interface due to cyclic reference. If that would be possible, we would be protected against name clashes. In the current implementation the names `Known` and `Unknown` can clash with enum name itself.

Current implementation uses Java 8 features (`Map.computeIfAbsent` and lambdas). I think, that this should be OK, but if not, it is possible to rewrite it in a Java 7 compatible code.

Probably not all cases handled, at least I did not test switch generation. If corresponding runtime changes will be merged, then CI would help with that.

Fixes the following https://ci.kaitai.io tests:
- EnumToIInvalid


generalmimon (changes requested):
@Mingun Thanks for suggesting a possible implementation. I don't like the fact that the `Known` and `Unknown` classes are nested in the interface, though. The main problem with this that which bothers me a lot is that it makes the access to concrete enum values (which is something user applications working with generated Java parsers will be probably doing most of the time when working with enums) very cumbersome and awkward because `.Known` always needs to be added (for example, <code>Enum0.Animal<ins><strong>.Known</strong></ins>.CAT</code> instead of `Enum0.Animal.CAT`), as is very visible in https://github.com/kaitai-io/kaitai_struct_tests/pull/125/files. And even users that don't care about unknown enum values at all would have to pay this "cost" of having `.Known` everywhere, which means they would be paying for something they don't use.

The structure suggested by @dgelessus in https://github.com/kaitai-io/kaitai_struct/issues/778#issuecomment-667416816 makes more sense to me overall. It doesn't change the way known enum values are accessed, which is good. And prefixing the enum name with `I` to name the interface that can represent both known and unknown case is elegant within reason, even though it's supposedly not idiomatic to start the interface names with `I` in Java (see https://www.baeldung.com/java-interface-naming-conventions#examples-of-incorrect-interface-naming) - it is only a convention in C#.

It unfortunately allows for a name collision with another type or enum whose name matches `i_{enum_name}` where `{enum_name}` is the name of an existing enum (related: https://github.com/kaitai-io/kaitai_struct/issues/86), but that's not really that much of a problem. We should just "teach" the compiler that defining an enum called `{enum_name}` reserves not only `{enum_name}` itself (you cannot define a type that is also called `{enum_name}`, although even this is currently unchecked by the compiler, as noted in https://github.com/kaitai-io/kaitai_struct/issues/1019#issuecomment-1503853469) but also `i_{enum_name}`. So if there is another type or enum called `i_{enum_name}`, the compiler should reject the .ksy spec with an error that this name is already reserved by the `{enum_name}` enum and another name must be used.

Another possible collision is that a user-specified `unknown_{enum_name}` type might collide with the auto-generated `Unknown{EnumName}` class, but as @dgelessus also pointed out, that class doesn't need to be `public`: if it isn't, the compiler could actually rename the class arbitrarily to avoid the name collision.

(Just to be clear, these strategies to combat name collision are just me thinking out loud, I definitely don't want any of this in this PR.)

But as you mentioned, your solution also doesn't have the advantage of avoiding name clashes:

> Unfortunately, it is not possible to generate enum what will implement nested interface due to cyclic reference. If that would be possible, we would be protected against name clashes. In the current implementation the names `Known` and `Unknown` can clash with enum name itself.

So I don't really see the benefit of the approach with nested classes. I'd really prefer the design proposed by @dgelessus in https://github.com/kaitai-io/kaitai_struct/issues/778#issuecomment-667416816.

But I like the code reuse mechanism that you used for `Unknown` (i.e. having the implementation in the runtime library so that it doesn't have to be repeated over and over again in the generated code).

Mingun:
Ok, I changed representation to that code, which is protected from any clashes in names ([proof](https://godbolt.org/z/jdqhf86eo)). Enum named `enum` would be generated as:
```java
// interface to abstract known and unknown values
interface IEnum extends IKaitaiEnum {
    // Storage for unknown values
    public static class Unknown extends IKaitaiEnum.Unknown implements IEnum {
        Unknown(long id) { super(id); }
        @Override
        public String toString() { return "Enum(" + this.id + ")"; }
    }
}
// Storage for known values
class Enum implements IEnum {
    VARIANT_1(1),
    VARIANT_2(2),
    VARIANT_3(3);

    private final long id;

    private static HashMap<Long, IEnum> variants = new HashMap<>(3);
    static {
        for (final Enum e : values()) {
            variants.put(e.id, e);
        }
    }

    public static IEnum byId(final long id) {
        return variants.computeIfAbsent(id, _id -> new IEnum.Unknown(id));
    }

    private Enum(long id) { this.id = id; }
    @Override
    public long id() { return this.id; }
}
```

generalmimon (approved):
@Mingun Thanks a lot for this valuable contribution, merging in!

shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala
```diff
@@ -966,23 +1020,32 @@ class JavaCompiler(typeProvider: ClassTypeProvider, config: RuntimeConfig)
...
     out.dec
+    out.puts("}") // for
     out.dec
-    out.puts("}")
-    out.puts(s"public static $enumClass byId(long id) { return byId.get(id); }")
+    out.puts("}") // static initializer
+    out.puts
+    out.puts(s"public static $enumIface byId(final long id) {")
+    out.inc
+    out.puts(s"return variants.computeIfAbsent(id, _id -> new $enumIface.Unknown(id));")
```
generalmimon, line 1036: Most of the time, my concerns about this line of code are probably not going to matter, so I merged it anyway. However, I actually have two concerns about this mutation of the hash map with enum entries done by the [`java.util.Map.computeIfAbsent()`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-) method:

1. The number of "unknown enum entries" is essentially unlimited (it's limited only by the range of the `long` data type, which is enormous). Once any given integer value is resolved, it gets stored in the hash map permanently and will prevent the `Unknown` object from being garbage-collected for the rest of the program execution, even if it hasn't been needed for a long time. This can be a problem for any long-running services that parse many different files - this behavior essentially becomes a memory leak / DoS vector.

   If [`WeakHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) was used instead, this problem would disappear. However, that would of course mean that the known and unknown entries would have to be put into different hash maps, because we don't want the known enum entries to be randomly garbage-collected at any time.

2. Although Kaitai Struct can only parse sequentially, there shouldn't be any reason why you couldn't parse different files of the same format from different threads. However, since this is a static non-thread-safe `HashMap` that is shared between threads, there could be a race condition if one thread mutates it while other threads read from it. We've had race conditions like this happen on a non-thread-safe `HashMap` in the compiler before: https://github.com/kaitai-io/kaitai_struct/issues/951#issuecomment-1981882454

   Of course, Java offers ways to deal with this problem too. From the [`WeakHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) documentation:
   
   > Like most collection classes, this class is not synchronized. A synchronized `WeakHashMap` may be constructed using the [`Collections.synchronizedMap`](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) method.

Given these concerns, I'm not convinced that caching the unknown entries at all is worth it, since this is purely an optimization that should have no impact on the results. For now, let me just remove it, and we can revisit this later if there's interest... If we do, we should ideally also have at least a basic benchmark, so that we know that it would make things better, not worse.

I feel like the `Unknown` class is basically a glorified boxed integer type like `Long`, and since it's very common in Java to create lots of boxed integers, it shouldn't be that much of an issue to create many `Unknown` objects either... Maybe reusing them if possible is better, but it's probably not the end of the world if we don't.

PR #290, opened 2024-03-27
Add missing translator tests for C++, Nim and Rust

Mingun:
## Do not merge yet, it is not finished. Not a draft because I want to run CI checks

I noticed that `TranslatorSpec.scala` does not test some languages, which seems to be oversight for Nim, and unfinished work for Rust.

Nim compiler seems to be production ready, so tests needs to be added.

Rust compiler is not finished yet, and tests should make finishing it easy.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala
@@ -413,6 +413,7 @@ class TranslatorSpec extends AnyFunSpec {
   describe("arrays") {
     describe("literals") {
       full("[0, 1, 100500]", CalcIntType, ArrayTypeInStream(CalcIntType), ResultMap(
+        CppCompiler -> "",
         CSharpCompiler -> "new List<int> { 0, 1, 100500 }",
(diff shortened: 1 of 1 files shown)


pombredanne:
Thanks for the work there. Rust would be awesome!

generalmimon:
@pombredanne:

> Rust would be awesome!

There is already quite decent support for Rust on unreleased `master` branches, as explained at https://app.gitter.im/#/room/!gxFYJoZuKcRouBfhlb:gitter.im/$U2qXwDnOd6HKOMzdHpcXehPmfg95pxgP4VA2D0CTtMk. So if you don't mind building the Kaitai Struct compiler from source (which is actually easy, as you can see at https://doc.kaitai.io/serialization.html#_building_the_compiler_from_source - just make sure to check out the `master` branch, not <code><del>serialization</del></code> as it says), you can already try it out now before a new version of Kaitai Struct is released. If you do, please give us feedback in https://github.com/kaitai-io/kaitai_struct/issues/22 or open issues in https://github.com/kaitai-io/kaitai_struct/issues/new for bugs/suggestions/questions.

PR #291, opened 2024-03-28
Use hex escape sequences instead of octal escape sequences

Mingun:
Octal escape sequences the least used form of escape sequences and hex supported almost everywhere. The only outsiders are [Java](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.6) (does not support `\xHH` form), [C++](https://en.cppreference.com/w/cpp/language/escape) (unlimited length, so have problems with followed digits) and [Rust](https://doc.rust-lang.org/reference/tokens.html#examples) (limited range of representable characters -- only `[0x00; 0x7f]`).

Seriously, who decided that using octal escape sequences by default was a good idea?


Mingun:
This is ready for review.

shared/src/main/scala/io/kaitai/struct/JSON.scala
```diff
@@ -28,7 +28,6 @@ object JSON extends CommonLiterals {
     }
   }
 
-  /** octal escapes (which [[translators.CommonLiterals.strLiteralGenericCC]] uses by default) are not allowed in JSON */
   override def strLiteralGenericCC(code: Char): String = strLiteralUnicode(code)
```
generalmimon, line 31: This comment should be updated, not deleted. The fact that we're overriding `strLiteralGenericCC` at all still deserves an explanation. Nothing has changed about that, because JSON allows neither octal nor hex escapes.

Mingun: Returned back with "hex escapes"

PR #293, opened 2024-03-29
Delete unused `FixedContentsUsingArrayByteLiteral` trait

Mingun:
Its usage was removed in 3fcc700017d4add59724338230604977952dbaec

As a next step need to delete method `ensureFixedContents` from runtime libraries as no longer used


generalmimon:
I've already done this on the `serialization` branch: https://github.com/kaitai-io/kaitai_struct_compiler/pull/255/commits/83b1fa74dfb32b7b8b9b6d8361a8ceef38394288

Mingun:
You yourself said that many small PRs are better than one large one with a thousand things. It seems to me that it is better to merge this one than to wait another N years when serialization will be recognized as ready enough for merge.

PR #297, opened 2024-04-07
C++: add a method to translator to get the raw string literal instead of std::string object

Mingun:
This method will be used in KST translator in https://github.com/kaitai-io/kaitai_struct_tests/pull/114


Mingun:
Superseded by #302

PR #299, opened 2024-04-13
Go: introduce special GoOutputWriter and move generation of error check to it

Mingun:
This will allow to override this method and do not generate check in KST when performing checks for returned errors using `asserts[i].exception` key.

This should be merged with this PR: https://github.com/kaitai-io/kaitai_struct_tests/pull/123


shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -15,7 +15,30 @@ sealed trait TranslatorResult
 case class ResultString(s: String) extends TranslatorResult
 case class ResultLocalVar(n: Int) extends TranslatorResult
 
-class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, importList: ImportList)
+class GoOutputWriter(indentStr: String) extends StringLanguageOutputWriter(indentStr) {
```
GreyCat, line 18: I'm very unsure if this is a way to go. Custom OutputWriters are supposed to be used to encompass very special logic: for example, something with very custom formatting (rather than traditional "keep indentation and go line-by-line").

Here, you use it to implement what can be implemented in any `Compiler` or `Translator` — i.e. pretty much vanilla output by calling `puts()`, `inc()`, `dec()`, etc.

If for some reason keeping it in GoTranslator does not work — let's discuss why and if we should move it into, for example, GoCompiler?

Mingun: The `GoOutputWriter` just add the customization point to intercept attempts to write the error check. Currently KS Translator [uses custom writer and heuristics](https://github.com/kaitai-io/kaitai_struct_tests/blob/ede1105eb313c222b7e1d9f1c357cfaac4b6a964/translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/GoSG.scala#L21-L24) to find and replace the returned error with `t.Fatal(err)` call.

This PR allows do the same but in a more elegant way, and also allow to completely skip the error check which is needed for the translator when the error is an expected result.

Probably it is possible to keep this logic in `GoTranslator` and make a subclass of the translator in KS Translator to use it [here](https://github.com/kaitai-io/kaitai_struct_tests/blob/ede1105eb313c222b7e1d9f1c357cfaac4b6a964/translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/GoSG.scala#L35).

generalmimon: I agree with @GreyCat that extending `StringLanguageOutputWriter` makes little sense in this case from the semantic point of view. I don't see why extending `GoTranslator` in the KST translator wouldn't work just as well.

Mingun: Ok, I rewrited https://github.com/kaitai-io/kaitai_struct_tests/pull/123 so this PR is no longer needed, unless you want to accept a couple of lines of documentation

PR #301, opened 2024-04-14
Fix problems with ownership in validation code

Mingun:
The actual fix contained in the first commit. The next commits just a minor style nits, which makes generated code looks better and gives more meaningful name to couple of constants.

The problem with validation code can be demonstrated with the following addition to `debug_array_user.ksy`:
```yaml
# --debug (or actually --no-auto-read) with arrays of user types requires
# special handling to avoid spoiling whole object due to exception handler.
meta:
  id: debug_array_user
  ks-debug: true
seq:
  - id: one_cat
    type: cat
    valid:
      expr: _._sizeof == 1
  - id: array_of_cats
    type: cat
    repeat: expr
    repeat-expr: 3
  - id: foo
    type: u1
    valid: # should pass
      expr: _ == 1
  - id: bar
    type: s2le
    valid: # there's actually -190 in the file
      expr: _ < -190 or _ > -190
types:
  cat:
    seq:
      - id: meow
        type: u1
```

I just copy some checks from the `valid_expr_fail.ksy` test to have them in one file so it easy to check what's changed while working on another issue. I noticed, that check for `one_cat` generates incorrect code for C++11: `one_cat()` getter returns `cat_t*`, but the result is stored in `std::unique_ptr<cat_t>`. This
- will not compile, because [constructor](https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr) is explicit
- even if compiled will lead to double-free problem


Mingun:
Also, `validation_expr_error` exception in C++ is dangerous because it store [reference](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/084efdd08ed174b6f916df0a71a4b7d948f3f1a8/kaitai/exceptions.h#L219) to the stack variable `_` (and temporary variable before the fix), but this is another problem. Other exceptions also store references and I think they reference to the stack variables as well.

PR #302, opened 2024-04-15
Fix incorrect encoding literals -- escape them

Mingun:
Due to C++ it was not possible to call `doStringLiteral` in `BaseTranslator`, because in C++ this method returns `std::string` object instead of a `"..."` literal.

Fixes the following KST tests:
- str_encodings_escaping_enc
- str_encodings_escaping_to_s

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala
@@ -404,6 +404,21 @@ class ExpressionsSpec extends AnyFunSpec {
       Expressions.parse("foo.bar") should be (Attribute(Name(identifier("foo")),identifier("bar")))
     }
 
+    describe("strings") {
+      it("single-quoted") {
+        // \" -> \"
+        // \\ -> \\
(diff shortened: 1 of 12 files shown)


shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala
```diff
@@ -316,7 +316,7 @@ class GoTranslator(out: StringLanguageOutputWriter, provider: TypeProvider, impo
             importList.add(importName)
             outVarCheckRes(s"kaitai.BytesToStr($bytesExpr, $decoderSrc.NewDecoder())")
           case None =>
-            throw new RuntimeException(s"encoding '$encStr' in not supported in Go")
+            throw new RuntimeException(s"encoding '${doStringLiteral(encStr)}' in not supported in Go")
```
generalmimon, line 319: The `doStringLiteral` is not needed here. This is an exception thrown by KSC that stops the compilation, it is not interpolated into the generated code.

Mingun: Yes, that was oversight from my side. Fixed now

generalmimon (approved):
@Mingun Thanks!

PR #303, opened 2024-04-17
Fix type and enum leaking from one imported KSY to another

### LINKED ISSUE #534: Imports "leak" across to other imported files
Using `kaitai-struct-compiler 0.9-SNAPSHOT501381150`.

It seems that specs imported by one spec are also implicitly imported for all other specs. This can lead to an imported spec compiling even though it doesn't import all needed types (because the only places that use it happen to also import the necessary types). Short example:

`one.ksy`:

```yaml
meta:
  id: one
seq:
  - id: something
    size: 4
```

`two.ksy`:

```yaml
meta:
  id: two
  # Note: missing import of `one`
seq:
  - id: one_thing
    type: one # Note: `one` is not imported in this spec, so this should cause a compile error
```

`toplevel.ksy`:

```yaml
meta:
  id: toplevel
  imports:
    - one
    - two
seq:
  - id: another_one
    type: one
  - id: another_two
    type: two
```

`one.ksy` is valid. `two.ksy` is invalid (because it uses `one` without importing it) and does not compile. `toplevel.ksy` itself is valid, but it imports the invalid `two.ksy`. Despite this, compiling `toplevel.ksy` does not produce an error, because `toplevel`'s import of `one` leaks into `two`. (If you remove the usage of `one` from `toplevel.ksy`, it will fail to compile, because then `two` cannot find the required `one` type anymore.)

Real example: https://github.com/kaitai-io/kaitai_struct_formats/pull/126#pullrequestreview-213104990

Mingun:
This PR implements solution from [my comment](https://github.com/kaitai-io/kaitai_struct/issues/534#issuecomment-2059567168). Each top-level `ClassSpec` now have a list of other top-level `ClassSpec`s which is explicitly import it. When resolve types and enums now checked if spec with the type/enum was explicitly imported or not and if not, it does not resolves.

Related links:
- https://github.com/kaitai-io/kaitai_struct_compiler/pull/303 -- this PR
- https://github.com/kaitai-io/kaitai_struct_tests/pull/124 -- tests
- Closes https://github.com/kaitai-io/kaitai_struct/issues/534

### DIFF
--- a/js/src/main/scala/io/kaitai/struct/format/JavaScriptClassSpecs.scala
+++ b/js/src/main/scala/io/kaitai/struct/format/JavaScriptClassSpecs.scala
@@ -13,9 +13,9 @@ class JavaScriptClassSpecs(importer: JavaScriptImporter, firstSpec: ClassSpec)
   val MODE_REL = "rel"
   val MODE_ABS = "abs"
 
-  override def importRelative(name: String, path: List[String], inFile: Option[String]): Future[Option[ClassSpec]] =
+  override def importRelative(name: String, path: List[String], inFile: String): Future[Option[ClassSpec]] =
     doImport(name, path, MODE_REL)
-  override def importAbsolute(name: String, path: List[String], inFile: Option[String]): Future[Option[ClassSpec]] =
+  override def importAbsolute(name: String, path: List[String], inFile: String): Future[Option[ClassSpec]] =
     doImport(name, path, MODE_ABS)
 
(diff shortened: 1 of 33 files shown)


Mingun:
Rebased on top of #310 (which includes #309)and test set updated to have enums in value instances. With this PR errors reported correctly, but strangely only the one (for seq, but not for value instance, when file contains both error as in the tests PR), even when I apply #313

PR #304, opened 2024-04-18
Correctly strip Windows paths from error messages

Mingun:
This PR correctly strips `..\tests\formats_err\` string from the error messages on Windows, while still strips `../tests/formats_err/` on \*nix.

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/ErrorMessagesSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/ErrorMessagesSpec.scala
@@ -14,12 +14,12 @@ class ErrorMessagesSpec extends AnyFunSuite with SimpleMatchers {
   // version info
   KSVersion.current = Version.version
 
-  val FORMATS_ERR_DIR = "../tests/formats_err"
+  val FORMATS_ERR_DIR = new File("../tests/formats_err")
   val CHARSET_UTF8 = Charset.forName("UTF-8")
   val DEFAULT_CONFIG = CLIConfig()
 
-  def getExpected(fn: String): List[String] = {
-    val fis = new FileInputStream(fn)
+  def getExpected(f: File): List[String] = {
+    val fis = new FileInputStream(f)
     val isr = new InputStreamReader(fis, CHARSET_UTF8)
     val br = new BufferedReader(isr)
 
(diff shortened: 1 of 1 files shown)


jvm/src/test/scala/io/kaitai/struct/ErrorMessagesSpec.scala
```diff
@@ -38,23 +38,22 @@ class ErrorMessagesSpec extends AnyFunSuite with SimpleMatchers {
   def testOne(f: File): Unit = {
     val fileName = f.getName
     val testName = fileName.stripSuffix(".ksy")
-    val fn = FORMATS_ERR_DIR + "/" + fileName
     test(testName) {
-      val expected = getExpected(fn)
-      val (_, problems) = JavaKSYParser.localFileToSpecs(fn, DEFAULT_CONFIG)
+      val expected = getExpected(f)
+      val (_, problems) = JavaKSYParser.localFileToSpecs(f.getPath(), DEFAULT_CONFIG)
```
Mingun, line 43: Honestly, I think it will be meaningful to pass `File` object to `localFileToSpecs`, but that method is called also in test generator in https://github.com/kaitai-io/kaitai_struct_tests repository, so I doesn't touch it yet.

Mingun:
What in this PR contradicts the information in https://github.com/kaitai-io/kaitai_struct_tests/pull/124#discussion_r1569683734? If I understand correctly, it is fine to have platform-specific delimiters in paths in error messages. This PR just create the same platform-specific path and strip it from the output. What's wrong with that approach? What bug you mean? Do you have a link to it? I saw https://github.com/kaitai-io/kaitai_struct/issues/507 earlier and having platform-specific delimiters was found to be expected behavior, no?

Even if you think that some bug is exist, it is not the reason to make life harder. If you find the time to fix that bug, you also will able to fix that place too, because you will see new failing tests.

Mingun:
I'm afraid, that the current solution is the best way to dealing with paths in the most agnostic way. The platform-specific path always could be present in the error message in case of non-existing imports:
```
[info] - meta_imports_rel2 *** FAILED ***
[info]   [meta_imports_rel_unknown.ksy: /meta/imports/0:
[info]          error: ..\tests\formats_err\unknown_relative_name.ksy (╨Э╨╡ ╤Г╨┤╨░╨╡╤В╤Б╤П ╨╜╨░╨╣╤В╨╕ ╤Г╨║╨░╨╖╨░╨╜╨╜╤Л╨╣ ╤Д╨░╨╣╨╗)
[info]   ]
[info]     did not equal
[info]   [meta_imports_rel_unknown.ksy: /meta/imports/0:
[info]          error: unknown_relative_name.ksy (No such file or directory)
[info]   ] (SimpleMatchers.scala:34)
```
The `..\tests\formats_err\unknown_relative_name.ksy (Не удается найти указанный файл)` is a message from `FileNotFoundException` which is thrown here and in the general can contains arbitrary string:
https://github.com/kaitai-io/kaitai_struct_compiler/blob/2191070c7a8a7e0c19b01ca79f5616437cc7bc6f/jvm/src/main/scala/io/kaitai/struct/formats/JavaKSYParser.scala#L39-L46

Current implementation contains OS-dependent path. It is possible to hack the error message, replace that path and rethrow the exception, but it seems too hacky.

I think, that instead of try to keep original paths in the ksc output, ksc can return a dictionary with mapping of provided paths to the actually used strings in messages. That way utilities which want to index JSON with paths should get a normalized path from the dictionary and use it.

PR #305, opened 2024-04-29
Julia

rystidia:
Extend KSC to support the Julia programming language as a target language. 

This PR addresses issue #1109


shared/src/main/scala/io/kaitai/struct/translators/JuliaTranslator.scala
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  override def doInternalName(id: Identifier): String =
+    s"this.${JuliaCompiler.publicMemberName(id)}"
+
+  override def doEnumByLabel(enumSpec: EnumSpec, label: String): String = {
+    val isExternal = enumSpec.isExternal(provider.nowClass)
+    if (isExternal) {
+      importList.add(s"import ${JuliaCompiler.type2class(enumSpec.name.head)}Module")
```
GreyCat, line 78: This `Module` suffix seems to be quite prominent in every single invocation to the classes, as far as I can tell. Please consider if we can add it to `JuliaCompiler.type2class`, rather than copy-pasting that `Module` constant everywhere?

Even if `type2class` won't work (e.g. if I overlook something and it's indeed needed in its base form), maybe consider doing another offspin like `JulaCompiler.type2module` or something like that.

rystidia: Introduced `JulaCompiler.type2module`

PR #308, opened 2024-09-15
Fix a number of errors in loops

Mingun:
This PR the result of revision of how generation of repeated attributes is implemented due to which a number of bugs was found and fixed:
- `TypeValidator` incorrectly define `_` variable for all kinds of loops, but it is generated only for `repeat-until` loops
- fixes https://github.com/kaitai-io/kaitai_struct/issues/958, closes #245
- C#, Go, Java, JavaScript was calculate count of repetitions of `repeat-expr` loops on each iteration, whereas other languages performs such calculations only once. Now they are also calculate count only once
- Now Perl defines `_index` variable in `repeat: eos` loops
- Go and Rust was incorrectly used root IO object instead of current IO object of the attribute. The difference should be noticable when `io: ...` is used

This PR supersedes #234

Also, some refactoring was done: removed unused parameters and documented methods that generates loops. 

Besides, `_currentIteratorType` was renamed to `_lastParsedType` (this is the variable that holds type of the `_` variable in expression language) because it is used also in `valid: expr` key to refer to the element that just was parsed.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassTypeProvider.scala
@@ -11,7 +11,16 @@ class ClassTypeProvider(classSpecs: ClassSpecs, var topClass: ClassSpec) extends
   var nowClass = topClass
   val allClasses: ClassSpecs = classSpecs
 
+  /**
+    * Type of the `_` variable in the expression. That variable is defined in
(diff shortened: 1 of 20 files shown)


transverberate:
Thank you for fixing this and closing the PR. I apologize for not following up on my PR last year. I'm happy we have this consistent behavior across languages.

PR #315, opened 2024-12-16
Fix operator precedence handling across target languages

AkiSakurai:
- Adds proper operator precedence tables for each target language and updates comparison operator handling to respect precedence rules.
- Fixes issues with bitwise operations and comparisons in Go, Lua, Perl, Ruby and other languages.


generalmimon:
Related: https://github.com/kaitai-io/kaitai_struct/issues/69

generalmimon (approved):
@AkiSakurai Thanks! I've extended this to cover even more cases, so now now I'm not aware of any remaining cases of missing parentheses.

PR #319, opened 2025-07-13
 feat(python): add full type annotations support with default enabled

mblsha:
This commit implements comprehensive Python type annotation support for the Kaitai Struct compiler, making it the default behavior for modern Python development while maintaining backward compatibility.

- **Type Annotations by Default**: Python code now generates with full type annotations enabled by default
- **Comprehensive Coverage**: Type annotations for constructors, methods, properties, and all data types
- **Modern Python Support**: Imports typing module and uses modern annotation syntax
- **Backward Compatibility**: New --no-python-type-annotations flag to disable when needed

- Added pythonTypeAnnotations: Boolean = true to RuntimeConfig (default: true)
- Created comprehensive kaitaiTypeToPythonType() mapping function
- Enhanced constructor with full parameter and instance variable annotations
- Added return type annotations for all methods (_read, properties, __repr__)
- Conditional typing imports only when annotations enabled
- Removed "# type: ignore" comment when annotations are present

**With type annotations (default):**
```python
from typing import Any, List, Optional, Union

def __init__(self, _io: 'KaitaiStream', _parent: 'KaitaiStruct' = None,
             _root: Optional['KaitaiStruct'] = None) -> None:
    self._io: 'KaitaiStream' = _io
    self._parent: 'KaitaiStruct' = _parent

@property
def calculated_value(self) -> int:
    return self.magic * 2
```

**Without annotations (--no-python-type-annotations):**
```python

def __init__(self, _io, _parent=None, _root=None):
    self._io = _io
    self._parent = _parent
```

- `--python-type-annotations`: Explicitly enable (now redundant but supported)
- `--no-python-type-annotations`: Disable type annotations
- Default behavior: Type annotations enabled

### DIFF
--- a/jvm/src/main/scala/io/kaitai/struct/JavaMain.scala
+++ b/jvm/src/main/scala/io/kaitai/struct/JavaMain.scala
@@ -117,6 +117,14 @@ object JavaMain {
         c.copy(runtime = c.runtime.copy(pythonPackage = x))
       } text("Python package (Python only, default: root package)")
 
+      opt[Unit]("python-type-annotations") action { (x, c) =>
+        c.copy(runtime = c.runtime.copy(pythonTypeAnnotations = true))
+      } text("generate Python type annotations (Python only, default: true)")
+
+      opt[Unit]("no-python-type-annotations") action { (x, c) =>
+        c.copy(runtime = c.runtime.copy(pythonTypeAnnotations = false))
+      } text("disable Python type annotations (Python only)")
+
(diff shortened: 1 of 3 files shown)


EmperorArthur:
@mblsha I've rebased your work on top of master here:  https://github.com/EmperorArthur/kaitai_struct_compiler/tree/python_type_annotations

EmperorArthur:
I have updated my branch to work with `--read-write`.

PR #320, opened 2025-08-18
Fix Python reserved keyword escaping in code generation

andrewmonostate:
### Problem
The Kaitai Struct compiler generates invalid Python code when field names use Python reserved keywords. For example, a field named `class` generates `self.class = ...`, which causes a `SyntaxError` because `class` is a reserved keyword in Python.

### Real-world Impact
This issue affects real format specifications, notably the OpenPGP message format which contains a field named `class`. The generated Python parsers cannot be imported due to syntax errors.

### Solution
This PR adds proper escaping of Python reserved keywords by appending an underscore, following PEP 8 conventions:
- `class` → `class_`
- `def` → `def_`
- `if` → `if_`
- etc.

### Changes Made

1. **Added reserved keyword detection** (`PythonCompiler.scala`):
   - Added `PYTHON_RESERVED_WORDS` set containing all Python reserved keywords
   - Added `escapePythonKeyword()` function to check and escape keywords
   - Modified `idToStr()` to apply escaping for `NamedIdentifier` and `InstanceIdentifier`

2. **Added comprehensive tests** (`PythonCompilerSpec.scala`):
   - Tests for common reserved keywords (`class`, `def`, `if`, `lambda`, etc.)
   - Tests for async/await keywords
   - Tests that non-reserved words are not escaped
   - Tests for instance identifiers with reserved keywords
   - All 12 tests pass ✅

### Test Results
```
[info] PythonCompilerSpec:
[info] Run completed in 233 milliseconds.
[info] Total number of tests run: 12
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 12, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

### Example
**Before (causes SyntaxError):**
```python
self.class = self._io.read_u1()
```

**After (valid Python):**
```python
self.class_ = self._io.read_u1()
```

### Compatibility
- Works with all Python versions (reserved keywords have been stable)
- Follows Python's PEP 8 naming conventions
- Non-breaking change for existing valid field names

### DIFF
--- a/jvm/src/test/scala/io/kaitai/struct/languages/PythonCompilerSpec.scala
+++ b/jvm/src/test/scala/io/kaitai/struct/languages/PythonCompilerSpec.scala
@@ -1,0 +1,83 @@
+package io.kaitai.struct.languages
+
+import io.kaitai.struct.format.{Identifier, NamedIdentifier, InstanceIdentifier}
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers._
+
+class PythonCompilerSpec extends AnyFunSpec {
+  describe("PythonCompiler") {
+    describe("reserved keyword escaping") {
+      it("should escape Python reserved keyword 'class'") {
+        val id = NamedIdentifier("class")
+        val result = PythonCompiler.idToStr(id)
+        result should be("class_")
+      }
+
+      it("should escape Python reserved keyword 'def'") {
+        val id = NamedIdentifier("def")
+        val result = PythonCompiler.idToStr(id)
+        result should be("def_")
+      }
+
+      it("should escape Python reserved keyword 'if'") {
+        val id = NamedIdentifier("if")
+        val result = PythonCompiler.idToStr(id)
+        result should be("if_")
+      }
+
+      it("should escape Python reserved keyword 'lambda'") {
+        val id = NamedIdentifier("lambda")
+        val result = PythonCompiler.idToStr(id)
+        result should be("lambda_")
+      }
+
+      it("should escape Python reserved keyword 'return'") {
+        val id = NamedIdentifier("return")
+        val result = PythonCompiler.idToStr(id)
+        result should be("return_")
+      }
+
+      it("should escape Python reserved keyword 'async'") {
+        val id = NamedIdentifier("async")
+        val result = PythonCompiler.idToStr(id)
+        result should be("async_")
+      }
+
+      it("should escape Python reserved keyword 'await'") {
+        val id = NamedIdentifier("await")
+        val result = PythonCompiler.idToStr(id)
+        result should be("await_")
+      }
+
+      it("should not escape non-reserved word 'class_name'") {
+        val id = NamedIdentifier("class_name")
+        val result = PythonCompiler.idToStr(id)
+        result should be("class_name")
+      }
+
+      it("should not escape non-reserved word 'my_field'") {
+        val id = NamedIdentifier("my_field")
+        val result = PythonCompiler.idToStr(id)
+        result should be("my_field")
+      }
+
+      it("should escape reserved keyword in InstanceIdentifier") {
+        val id = InstanceIdentifier("class")
+        val result = PythonCompiler.idToStr(id)
+        result should be("_m_class_")
+      }
+
+      it("should handle privateMemberName with reserved keyword") {
+        val id = NamedIdentifier("class")
+        val result = PythonCompiler.privateMemberName(id)
+        result should be("self.class_")
+      }
+
+      it("should handle privateMemberName with normal identifier") {
+        val id = NamedIdentifier("normal_field")
+        val result = PythonCompiler.privateMemberName(id)
+        result should be("self.normal_field")
+      }
+    }
+  }
+}
--- a/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala
@@ -546,12 +546,30 @@ object PythonCompiler extends LanguageCompilerStatic
     config: RuntimeConfig
   ): LanguageCompiler = new PythonCompiler(tp, config)
 
+  // Python reserved keywords that need to be escaped
+  // https://docs.python.org/3/reference/lexical_analysis.html#keywords
+  val PYTHON_RESERVED_WORDS: Set[String] = Set(
+    "False", "None", "True", "and", "as", "assert", "async", "await",
+    "break", "class", "continue", "def", "del", "elif", "else", "except",
+    "finally", "for", "from", "global", "if", "import", "in", "is",
+    "lambda", "nonlocal", "not", "or", "pass", "raise", "return",
+    "try", "while", "with", "yield"
+  )
+
+  def escapePythonKeyword(name: String): String = {
+    if (PYTHON_RESERVED_WORDS.contains(name)) {
+      name + "_"
+    } else {
+      name
+    }
+  }
+
   def idToStr(id: Identifier): String =
     id match {
       case SpecialIdentifier(name) => name
-      case NamedIdentifier(name) => name
+      case NamedIdentifier(name) => escapePythonKeyword(name)
       case NumberedIdentifier(idx) => s"_${NumberedIdentifier.TEMPLATE}$idx"
-      case InstanceIdentifier(name) => s"_m_$name"
+      case InstanceIdentifier(name) => s"_m_${escapePythonKeyword(name)}"
       case RawIdentifier(innerId) => s"_raw_${idToStr(innerId)}"
     }
 


Mingun:
> @Mingun I'm not convinced that numeric suffixes are a good idea,

Using numerical suffixes is not necessary, but it is the simplest scalable solution with clear semantics and predictable rules.

> because you inevitably make the name translation of one attribute name dependent on what the other attributes are named.

This is the only possible solution if you want to prevent name conflicts. Conflict implies you need to know about other names in the same scope to ensure you prevent it

> This implies that if the .ksy spec is changed in a way that some of the other attributes are removed/renamed, it may influence this attribute, even though it was not touched.

Of course, but this only will happened with attributes, which already requires some non-trivial name translation (mostly attributes with reserved names). The number of such attributes, however, is quite small, because there are usually few reserved words and it is unlikely that other attributes will also appear next to them, whose names match the name of the reserved attribute after its trivial transformation. Anyway, if you remove / rename something, you already make incompatible changes and it is expected, that dependers should be also updated. Use semver and everything will be fine.

> Another problem is that with numeric suffixes, you cannot tell from the .ksy specification how conflicting names will be numbered. For example, if you have `foo_bar_` and `foo__bar`, will it be `fooBar` and `fooBar1` respectively, or the other way around? You can issue warnings about this, which mitigates this problem somewhat, but I don't think you should be forced to study warnings just to learn what today's mapping from .ksy names to generated code names is.

You really does not have any other option. The only advice is to try not to name names too similar in one scope (differing only in the number of underscores, for example).

> I also don't think we need to be linguistic purists - we don't have to support everything. I don't think a .ksy spec with two names in the same scope that differ only in underscores follows good practices (and also if we reject it, I'd argue it's unlikely to affect many people), so I definitely agree with @GreyCat on stricter naming rules. If there's a need for similar names, why couldn't the author of the .ksy specification add a numerical suffix themselves (e.g. `id: foo_bar`, `id: foo_bar1`)? This would avoid the problems I have described.

You lose sight of the fact that the main problem is that the converted names may match the existing ones, even if everything is fine in your KSY. And you need to solve this problem, because it is impossible to prohibit all words reserved in all languages or otherwise unacceptable. An example is Java. You can name the type in KSY as `class`. Working with this in Java will be unpleasant, but possible. But you cannot name the attribute `class` because, according to the conversion rules for Java, this will turn into `getClass()`, which is name of the special method exists in all objects. You naturally cannot foresee all similar cases in all languages.

Mediocre software distinguishes from good software how the corner cases processed. Don't be mediocre. As you yourself said, we don't have to support everything. But we can make things more end-user friendly. The user does not need to think why he cannot use the `as` word as attribute name when generating Java code, because somewhere in Python for which he was not going to and never will be, this is a reserved word. In Java, it will always turn into either `As`, `getAs` or `setAs` and I don't see any problems with that.

andrewmonostate:
I want to address the whole “underscore vs numeric suffix” debate in detail, because frankly this is seems like unreal to me, very over-engineered and it distracts from the actual bug we need to solve, which is actual people's code not compiling.

1. The bug is real and already proven.
Generated Python code with reserved keywords (class, def, etc.) simply doesn’t import — it throws a SyntaxError. I already provided a compiler-level unit test in my PR, and now there’s a duplicated integration test in the main suite. Great, the bug is confirmed. The conversation should have stopped there: fix it with the most idiomatic, Python-appropriate solution.

2. Underscore is the Pythonic solution.
This isn’t arbitrary. PEP8, the Python stdlib, and virtually every OSS project uses a trailing underscore to escape reserved keywords. Example straight from the docs: class_. This is clear, readable, predictable, and already familiar to every Python developer. Numeric suffixes (class2, class3) are not idiomatic, harder to read, and make code feel generated and clunky.

3. The “but what if foo_bar_ and foo__bar collide” argument is absurd.
Yes, in pathological cases you can construct names that are weird after mangling. But the same is true in every language. At some point, if you’re authoring a .ksy with foo_bar_ and foo__bar in the same scope, you’re already writing bad specs. It’s not Kaitai’s job to enable every self-contradictory naming scheme under the sun. This is no different from C disallowing two functions that only differ by case on Windows, or Java banning identifiers that collide with getters. You set reasonable rules, not try to support nonsense.

4. “Numeric suffixes scale better” is a non-argument.
The proposed numbering system is worse because:
	•	It makes name resolution dependent on what else is in the scope (brittle).
	•	It produces unpredictable mappings (class might become class2 in one version, class3 in another if something else changes).
	•	It’s not transparent from the .ksy spec — you can’t tell what the final code will look like without generating it.
	•	It breaks the principle of least surprise for Python developers.

This “solution” actually introduces instability into downstream code. You edit your .ksy slightly and suddenly different attributes get renumbered. That’s not robustness, that’s chaos at your fingertips!

5. “We must preserve language-agnosticism” is absurd.
Kaitai already applies language-specific rules (camelCase, getX in Java, etc.). That’s the entire point of translators. Pretending that the Python backend can’t adopt Pythonic conventions because “language agnosticity” is just moving the goalposts. By that logic, we shouldn’t be adding any reserved-word handling at all, because it’s inherently language-specific.

6. Practical reality beats theoretical puritans.
Right now, users cannot generate working Python code. That’s a real, blocking bug. The underscore solution fixes it immediately in a way that is familiar and natural. Numeric suffixing and endless debates about edge-case underscores solve nothing except giving contributors the feeling of pain and stalling an obvious fix.

Conclusion:
	•	Append _ to reserved keywords (PEP8 convention, stable, predictable).
	•	why overthink pathological naming collisions? They are rare, and stricter rules on the .ksy side are fine if needed.
	•	Stop chasing numeric suffix fantasies - they make code unpredictable and harder to maintain.
	•	Merge the tested, working fix, and then discuss incremental refinements if truly necessary.

Right now the project is burning contributor goodwill over pointless nitpicking instead of just shipping a fix that unblocks users. That’s the real risk here, not whether someone somewhere might define both foo_bar_ and foo__bar.

And by the way, I see uthis exact discussion has been going in circles since 2016 in issue #90. That’s 9 years of arguing over edge-cases while leaving the real bug unfixed. This is exactly why contributors get frustrated. The practical solution here is simple and Pythonic: underscore escape.

andrewmonostate:
If you guys need more details or proof of the error, see my original issue #1249. I provided a concrete Python reproduction showing how class generates invalid code (SyntaxError). Instead of treating that as a separate, actionable Python bug, it was closed and merged into #90 - a 9-year-old thread that’s been stuck in theoretical name-mangling debates since 2016 (2 world cups, 2 Olympic Games and a pandemic has passed and that issue is still alive!). 

That issue is exactly why nothing has moved: endless speculation about “perfect universal rules” while real users are left with broken generated code. It makes no sense to bury fresh, clear bug reports into decade-old design debates. The practical fix is simple (underscore escape) and already tested and cleanly implemented.

Mingun:
> I want to address the whole “underscore vs numeric suffix” debate

I think you got the substance of the debate wrong. There is no "undercore vs numeric suffix" debates. There is only a discussion of how to implement correctly, without introducing hidden bugs. For example, if you look at my suggestion carefully, you will see that the concrete language translator itself decides exactly how it will convert the original names into final ones. If you want, on the first attempt you can just add an underscore. If this name is free, it will be used. But if not, you'll have another go. Now you can add a number, add a number and underscore, underscore and number, or whatever you want.

Mingun:
PEP8 has concrete purpose: allow identifiers, that looks like keywords. It does not solve the name clashing problem, which translator should prevent when rename rules -- because if KSY model is checked and corrent, the translator must generate the correct code.

You trying to solve only the forbidden identifiers problem, but at the same time introduce another bug by creating a name clash problem. In my opinion, if you do something, you should do it right, i.e. solve that problem, especially when it was highlighted and the solution was suggested.

> makes downstream code unstable when .ksy specs evolve.

_Any_ changes makes downstream code unstable. You may rename `class` to `clazz` and then downstream code will stop to compiling / working correctly.

> No naming scheme can be bulletproof across all languages

That is not required. Read what I wrote several times: _each language_ should use its own naming scheme. This is what I am trying to convey and what @generalmimon disagrees with. They may be identical for different languages, but that is not required.

> That’s why Python should follow Pythonic rules, Java should follow Java rules, etc.

Yes. Yes! Use the idiomatic "to pythonic case + append underscore" on the first try, but use something else when this will not work. The code snippet, provided by me, only requires from you only to define rules what to do on each try.

> That produces names that are unpredictable and brittle: class could become class_ in one version and class1 in another if something else collides later.

They predictable. Just the numbers are the most predictable parts of names. `class -> class_` may only be changed to `class -> class1` if you add field `class_` to the KSY. Because explicit names always takes precedence. if the name `class_` already explicitly exists, don't you think it's strange to rename some other field to it?

> 4.	Scope creep vs fixing the bug.
> 5.	Complement, not replace.

Come on, this project is in a half-dead state. If you need the fix, maintain you own fork. There, if you allow yourself, you can implement any ad-hoc solutions, which works for you case but may break other cases that you are not interesting. But I feel that for shared project this will be totally wrong approach.

> Not stall the urgent bugfix for another abstract round of “what if” scenarios.

I would agree with that if the correct fix was not known, but it is not. And I as you may noticed by yourself, any "urgent" fixes should come to you own fork (of course, if under "urgent" you mean something other than "in this decade, maybe").

andrewmonostate:
@Mingun  I think you’re overstating the “clash problem” and missing the practical priority here.

	1.	PEP8 and reserved identifiers
PEP8 isn’t just about “looking like” keywords — it’s the de facto convention for escaping them. class → class_ is what every Python developer expects. That solves the real, immediate bug: generated code doesn’t even import. That’s priority #1.

	2.	“But you introduce a clash bug”
This is a hypothetical overreaction. If someone actually defines both class and class_ in the same .ksy, they’ve already written a bad spec. Kaitai already rejects or transforms other self-contradictory cases (like underscores collapsing in camelCase languages). It’s not the translator’s job to endlessly sanitize pathological naming collisions — it’s the spec author’s job to not create them. Right ? 

	3.	“Numbers are the most predictable part of names”
This is where we disagree. A user reading class_ knows immediately what happened: it was a reserved word, escaped Pythonically. A user reading class2 has no clue unless they reverse engineer the translator’s fallback logic. Worse: when .ksy evolves, numbers do shift around, which silently breaks downstream code. That’s not predictability, that’s instability disguised as determinism

	4.	“Just fork it if you need it”
With respect, that’s not how healthy OSS works. The point of this project is to provide working codegen for supported languages. Telling contributors to fork and patch themselves instead of merging a minimal, idiomatic bugfix is how projects rot. This is the definition of paralysis by analysis - endless “what if” debates while users are stuck with broken code.

	5.	Correct vs practical
You keep saying “the correct fix is known.” The reality is: no universal “correct” exists. Each language translator needs to follow its own idioms. For Python, the correct fix is underscore escaping. The “try underscore, then numbers, then something else” scheme is not more correct — it’s just more complex, less Pythonic, and introduces new fragility.

And just to be clear: we already had to patch around this bug in downstream repos — see trailofbits/polyfile#3431, trailofbits/polyfile#3432, and zbirenbaum/polyfile-weave#1. That’s duct tape to keep real projects working, not theory. The bug is blocking users today. Saying “just fork it” is not a serious answer - the whole point of Kaitai is to provide working, idiomatic generators. If the only way to get fixes in is by faking it downstream, that’s exactly how an OSS project would stop being useful.

--//--
TLDR
No human writing a spec - or machine for that matter - would ever deliberately create both class and class_ side by side. It’s a contrived edge case with no practical value, so what’s the point of blocking a real bugfix over it?

PR #335, opened 2026-04-16
Implement new `enums` syntax (enforced underlying `type`, `doc` / `doc-ref` support)

### LINKED ISSUE #92: ksc: out of range enums
C# and Java (possibly others) are using `s4` enums, but the compiler generates `u4` enums which are out-of-range if values above 2**31 used.

In C# we can choose the underlying type for enums by using `public enum LoadCommandType: uint` instead of `public enum LoadCommandType`.

### LINKED ISSUE #358: Documentation support for enums
It's probably a good idea to add documentation support for top-level enums. One of the approached suggested by @julie-bsx is:

```yaml
enums:
  animals:
    doc: |
      Animals you might find at the office.
    1: dogs
    2: cats
    3: birds
    4: programmers
```

Original discussion was in https://github.com/kaitai-io/kaitai_struct/issues/356

### LINKED ISSUE #862: C#: Invalid code for unsigned enums

### LINKED ISSUE #1288: New `enums` syntax to allow enum-level properties (`doc` / `doc-ref`, required underlying type)
Since the current `enums` syntax contains the enum values directly under the name of the enum, it doesn't allow for any "global" properties for the entire enum.

https://github.com/kaitai-io/kaitai_struct/issues/358 discusses that this prevents support for `doc` / `doc-ref` keys to document the enum (if we disregard the ugly hack of mixing `doc` and integer keys on the same YAML level, as shown in https://github.com/kaitai-io/kaitai_struct/issues/358#issue-296885472, which IMO shouldn't even be considered). These would be optional properties that have no impact on the overall behavior of the language, though.

What's more important is that in https://github.com/kaitai-io/kaitai_struct/issues/862#issuecomment-2241270048, @GreyCat pointed out that the "untyped enums" that we have right now are not a very good idea. Pretty much all statically typed target languages require or encourage specifying the underlying integer type that the enum is based on. Sometimes you can *technically* omit specifying it explicitly, but that just means that some underlying type will be decided by the target language instead, which is often not what we want.

For example, https://github.com/kaitai-io/kaitai_struct/issues/862 shows that C# uses `int` (a 32-bit signed integer type) by default, which effectively limits our C# enum support to signed 32-bit integers. In C++, the enum type can by default legally only represent integers with the minimum bit width that can fit all enum constants. If you try to convert a wider integer value to the enum type (https://github.com/kaitai-io/kaitai_struct/issues/778 is related), you are "loading out-of-range values to enums without fixed underlying type", which is undefined behavior - see https://github.com/kaitai-io/kaitai_struct/issues/959. The page https://pvs-studio.com/en/docs/warnings/v1016/ referenced in that issue has a good example:

> Example 3.
>
> ```cpp
> enum EN { low = 2, high = 4 }; // Uses 3 bits, range: [0; 7]
> EN a1 = static_cast<EN>(7);    // ok
> ```
>
> According to the standard, the underlying type for this enum is 'int'. Inside this type, the compiler uses the minimum width of the bit field that can fit all the values of enum constants.
>
> In this case, you will need at least 3 bits to fit all the values (2 = 0b010 and 4 = 0b100), so an EN variable can fit numbers from 0 (0b000) to 7 (0b111) inclusively. The number 8 already occupies four bits (0b1000), so it no longer fits in the EN type:
>
> ```cpp
> EN a2 = static_cast<EN>(8);    // UB
> ```
>
> UndefinedBehaviorSanitizer also finds an error in this example: https://godbolt.org/z/GGYo7z.

Therefore, "untyped enums" are not really a thing - the compiler should always know the exact underlying type of each enum and declare it in the generated code, if the target language supports that. Another reason that illustrates why this is important is the `to_i` method of enums, see the [User Guide](https://doc.kaitai.io/user_guide.html#enum-methods):

> #### 6.4.5. Enums
>
> | Method name | Return type | Description |
> |:---|:---|:---|
> | `to_i` | Integer | Converts an enum into the corresponding integer representation |

It says that the return type is "integer", but that's not specific enough - the compiler must know the exact type. In theory, this should be the enum's underlying type, but since the compiler doesn't know that at the moment, it uses `CalcIntType`, i.e. signed 32-bit integer:

[`TypeDetector.scala:220-222`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/3270e5fa9a582138fb01bb09fda6be0171710db8/shared/src/main/scala/io/kaitai/struct/translators/TypeDetector.scala#L220-L222)

```scala
      case et: EnumType =>
        attr.name match {
          case "to_i" => CalcIntType
```

So in statically typed languages, the result of the `enum.to_i` method is currently truncated to a signed 32-bit integer, regardless of the actual underlying type.

---

Therefore, I propose a new syntax for enum definitions (which combines @GreyCat's suggestions in https://github.com/kaitai-io/kaitai_struct/issues/358#issuecomment-365688068 and https://github.com/kaitai-io/kaitai_struct/issues/862#issuecomment-2241270048):

```ksy
enums:
  animals:
    doc: Animals you might find at the office.
    type: u4
    values:
      1: dogs
      2: cats
```

Obviously, the enum values themselves need to be moved to the `values` key to allow specifying other enum properties. That's a relatively trivial change. A bigger change is the new `type` key, which will be **REQUIRED** and will explicitly declare the enum's underlying type.

To address @GreyCat's concerns in https://github.com/kaitai-io/kaitai_struct/issues/862#issuecomment-2241270048, the underlying type will be enforced. First, this means that only integer constants representable by the underlying type will be allowed in the enum definition. Second, it will no longer be possible to directly convert an integer to an enum whose underlying type has a different width or signedness. So for the `u4`-based enum `animals` defined above, trying to use it on a `s4` field would be a compile-time error:

```ksy
seq:
  - id: animal
    type: s4  # error: 's4' does not match the underlying type 'u4' of the enum 'animals'
    enum: animals
```

Of course, this is a breaking change, but I believe it's the only future-proof solution that ensures consistent behavior across languages. What do you think?

generalmimon:
- Resolves https://github.com/kaitai-io/kaitai_struct/issues/1288
- Resolves https://github.com/kaitai-io/kaitai_struct/issues/358
- Fixes https://github.com/kaitai-io/kaitai_struct/issues/862
- Fixes https://github.com/kaitai-io/kaitai_struct/issues/92
- Closes https://github.com/kaitai-io/kaitai_struct_compiler/pull/248
- Solves https://github.com/kaitai-io/kaitai_struct/issues/959 for C++11 only (but not for C++98)

Related PR in the KST repo to update the test suite accordingly: https://github.com/kaitai-io/kaitai_struct_tests/pull/140

As explained in https://github.com/kaitai-io/kaitai_struct/issues/1288, this is a breaking change: the current `enums` syntax is declared old and unsupported ("legacy pre-v0.12 enum syntax"), and only the new syntax is accepted.

I can imagine that this will have a big impact, because suddenly almost everyone will have to update their .ksy specs. I suppose it would be possible to maintain support for the old syntax somehow, but that would significantly limit adoption of the new one... So I'm not sure if that's really what we want.

### DIFF
--- a/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
@@ -487,7 +487,7 @@ class ClassCompiler(
   }
 
   def compileEnum(curClass: ClassSpec, enumColl: EnumSpec): Unit =
-    lang.enumDeclaration(curClass.name, enumColl.name.last, enumColl.map.toSeq)
+    lang.enumDeclaration(curClass.name, enumColl.name.last, enumColl.doc, enumColl.map.toSeq)
 
   def compileClassDoc(curClass: ClassSpec): Unit = {
     if (!curClass.doc.isEmpty)
--- a/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
+++ b/shared/src/main/scala/io/kaitai/struct/ClassCompiler.scala
@@ -487,7 +487,7 @@ class ClassCompiler(
   }
 
   def compileEnum(curClass: ClassSpec, enumColl: EnumSpec): Unit =
-    lang.enumDeclaration(curClass.name, enumColl.name.last, enumColl.doc, enumColl.map.toSeq)
+    lang.enumDeclaration(curClass.name, enumColl.name.last, enumColl.map.toSeq, enumColl)
 
   def compileClassDoc(curClass: ClassSpec): Unit = {
     if (!curClass.doc.isEmpty)
--- a/shared/src/main/scala/io/kaitai/struct/RuntimeConfig.scala
+++ b/shared/src/main/scala/io/kaitai/struct/RuntimeConfig.scala
@@ -10,6 +10,14 @@ package io.kaitai.struct
   * @param useListInitializers If true, allows use of list initializers for
   *                            `std::vector` and `std::set`. Otherwise, throw a fatal
   *                            "not implemented" error.
+  * @param enumsFixedUnderlyingType If true, enums will be declared with fixed
+  * underlying type (see https://github.com/kaitai-io/kaitai_struct/issues/1288).
+  * If false, enums will be declared without a fixed underlying type. Enums without
+  * a fixed underlying type are best avoided, because casting an integer that falls
+  * outside the range of the enum values to the enum type is undefined behavior (see
+  * https://github.com/kaitai-io/kaitai_struct/issues/959). However, in C++98, this
+  * is the only supported option (see
+  * https://cppreference.com/cpp/language/enum#Unscoped_enumerations).
   * @param pointers Choose which style of pointers to use.
   */
 case class CppRuntimeConfig(
@@ -17,6 +25,7 @@ case class CppRuntimeConfig(
   usePragmaOnce: Boolean = false,
   stdStringFrontBack: Boolean = false,
   useListInitializers: Boolean = false,
+  enumsFixedUnderlyingType: Boolean = false,
   pointers: CppRuntimeConfig.Pointers = CppRuntimeConfig.RawPointers
 ) {
   /**
@@ -27,6 +36,7 @@ case class CppRuntimeConfig(
     usePragmaOnce = false,
     stdStringFrontBack = false,
     useListInitializers = false,
+    enumsFixedUnderlyingType = false,
     pointers = CppRuntimeConfig.RawPointers
   )
 
@@ -38,6 +48,7 @@ case class CppRuntimeConfig(
     usePragmaOnce = true,
     stdStringFrontBack = true,
     useListInitializers = true,
+    enumsFixedUnderlyingType = true,
     pointers = CppRuntimeConfig.UniqueAndRawPointers
   )
 }
--- a/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
@@ -54,6 +54,53 @@ object DataType {
       * The largest value of the given integer type
       */
     def max: BigInt
+    /**
+      * Render the integer type as a KSY-style "pure" type string (e.g. `u4`,
+      * `s1`, `b3`) without any serialization details, such as endianness.
+      * Intended for use in error messages.
+      */
+    def toPureTypeString: String
+  }
+  object IntType {
+    final val NUM_BITS_IN_BYTE = 8
+
+    /**
+      * Determine whether the two given integer types are equivalent, i.e.
+      * whether they have exactly the same range and can therefore represent the
+      * same set of integer values. Endianness is intentionally not compared,
+      * since it doesn't affect the set of representable values.
+      *
+      * [[CalcIntType]] represents an unspecified integer type, so we cannot
+      * assume that it's equivalent to anything, not even itself.
+      *
+      * This method is used to check whether an integer attribute can be
+      * converted to an enum type with a specific underlying type.
+      * https://github.com/kaitai-io/kaitai_struct/issues/1288 explains why only
+      * integer types equivalent to the underlying type are convertible to the
+      * enum type.
+      *
+      * @param a one integer type
+      * @param b other integer type
+      * @return `true` if the types are equivalent, `false` otherwise
+      */
+    def areEquivalent(a: IntType, b: IntType): Boolean =
+      (a, b) match {
+        case (Int1Type(sa), Int1Type(sb)) =>
+          sa == sb
+        case (IntMultiType(sa, wa, _), IntMultiType(sb, wb, _)) =>
+          sa == sb && wa == wb
+        case (BitsType(wa, _), BitsType(wb, _)) =>
+          wa == wb
+        // `u1` is equivalent to `b8`
+        case (Int1Type(false), BitsType(8, _)) => true
+        case (BitsType(8, _), Int1Type(false)) => true
+        // more generally, `u{N}` is equivalent to `b{8 * N}`
+        case (IntMultiType(false, byteWidth, _), BitsType(bitWidth, _)) =>
+          bitWidth == NUM_BITS_IN_BYTE * byteWidth.width
+        case (BitsType(bitWidth, _), IntMultiType(false, byteWidth, _)) =>
+          bitWidth == NUM_BITS_IN_BYTE * byteWidth.width
+        case _ => false
+      }
   }
   /**
     * In statically typed languages, [[CalcIntType]] is often a signed 32-bit
@@ -64,6 +111,7 @@ object DataType {
   case object CalcIntType extends IntType {
     override final def min: BigInt = ???
     override final def max: BigInt = ???
+    override def toPureTypeString: String = "<calc>"
   }
   case class Int1Type(signed: Boolean) extends IntType with ReadableType {
     override final def min: BigInt =
@@ -78,13 +126,12 @@ object DataType {
       } else {
         0xff
       }
+    override def toPureTypeString: String = if (signed) "s1" else "u1"
     override def apiCall(defEndian: Option[FixedEndian]): String = if (signed) "s1" else "u1"
   }
   case class IntMultiType(signed: Boolean, width: IntWidth, endian: Option[FixedEndian]) extends IntType with ReadableType {
-    private final def bitWidth: Int = {
-      val NUM_BITS_IN_BYTE = 8
-      NUM_BITS_IN_BYTE * width.width
-    }
+    private final def bitWidth: Int =
+      IntType.NUM_BITS_IN_BYTE * width.width
 
     override final def min: BigInt =
       if (signed) {
@@ -94,6 +141,8 @@ object DataType {
       }
     override final def max: BigInt =
       (BigInt(1) << (bitWidth - (if (signed) 1 else 0))) - 1
+    override def toPureTypeString: String =
+      s"${if (signed) 's' else 'u'}${width.width}"
 
     override def apiCall(defEndian: Option[FixedEndian]): String = {
       val ch1 = if (signed) 's' else 'u'
@@ -106,6 +155,8 @@ object DataType {
     override final def min: BigInt = 0
     override final def max: BigInt =
       (BigInt(1) << width) - 1
+    override def toPureTypeString: String =
+      s"b$width"
   }
 
   abstract class FloatType extends NumericType
--- a/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
@@ -45,12 +45,56 @@ object DataType {
   abstract sealed class NumericType extends DataType
   abstract sealed class BooleanType extends DataType
 
-  abstract sealed class IntType extends NumericType
-  case object CalcIntType extends IntType
+  abstract sealed class IntType extends NumericType {
+    /**
+      * The smallest value of the given integer type
+      */
+    def min: BigInt
+    /**
+      * The largest value of the given integer type
+      */
+    def max: BigInt
+  }
+  /**
+    * In statically typed languages, [[CalcIntType]] is often a signed 32-bit
+    * integer, but from the compiler's perspective, it represents an unspecified
+    * integer type. Therefore, we deliberately don't implement the [[min]] and
+    * [[max]] methods.
+    */
+  case object CalcIntType extends IntType {
+    override final def min: BigInt = ???
+    override final def max: BigInt = ???
+  }
   case class Int1Type(signed: Boolean) extends IntType with ReadableType {
+    override final def min: BigInt =
+      if (signed) {
+        Byte.MinValue
+      } else {
+        0
+      }
+    override final def max: BigInt =
+      if (signed) {
+        Byte.MaxValue
+      } else {
+        0xff
+      }
     override def apiCall(defEndian: Option[FixedEndian]): String = if (signed) "s1" else "u1"
   }
   case class IntMultiType(signed: Boolean, width: IntWidth, endian: Option[FixedEndian]) extends IntType with ReadableType {
+    private final def bitWidth: Int = {
+      val NUM_BITS_IN_BYTE = 8
+      NUM_BITS_IN_BYTE * width.width
+    }
+
+    override final def min: BigInt =
+      if (signed) {
+        -(BigInt(1) << (bitWidth - 1))
+      } else {
+        0
+      }
+    override final def max: BigInt =
+      (BigInt(1) << (bitWidth - (if (signed) 1 else 0))) - 1
+
     override def apiCall(defEndian: Option[FixedEndian]): String = {
       val ch1 = if (signed) 's' else 'u'
       val finalEnd = endian.orElse(defEndian)
@@ -58,7 +102,11 @@ object DataType {
     }
   }
   case class BitsType1(bitEndian: BitEndianness) extends BooleanType
-  case class BitsType(width: Int, bitEndian: BitEndianness) extends IntType
+  case class BitsType(width: Int, bitEndian: BitEndianness) extends IntType {
+    override final def min: BigInt = 0
+    override final def max: BigInt =
+      (BigInt(1) << width) - 1
+  }
 
   abstract class FloatType extends NumericType
   case object CalcFloatType extends FloatType
--- a/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
+++ b/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala
@@ -60,47 +60,18 @@ object DataType {
       * Intended for use in error messages.
       */
     def toPureTypeString: String
+
+    final def subsetOf(other: IntType): Boolean = {
+      // This implementation might be slightly inefficient (compared to pattern
+      // matching with a few simple comparisons) because it involves calculating
+      // large `BigInt` values, but it's guaranteed to give the correct result.
+      // That's because it directly tests the "subset of" relationship by
+      // comparing the `min` and `max` bounds.
+      min >= other.min && max <= other.max
+    }
   }
   object IntType {
     final val NUM_BITS_IN_BYTE = 8
-
-    /**
-      * Determine whether the two given integer types are equivalent, i.e.
-      * whether they have exactly the same range and can therefore represent the
-      * same set of integer values. Endianness is intentionally not compared,
-      * since it doesn't affect the set of representable values.
-      *
-      * [[CalcIntType]] represents an unspecified integer type, so we cannot
-      * assume that it's equivalent to anything, not even itself.
-      *
-      * This method is used to check whether an integer attribute can be
-      * converted to an enum type with a specific underlying type.
-      * https://github.com/kaitai-io/kaitai_struct/issues/1288 explains why only
-      * integer types equivalent to the underlying type are convertible to the
-      * enum type.
-      *
-      * @param a one integer type
-      * @param b other integer type
-      * @return `true` if the types are equivalent, `false` otherwise
-      */
-    def areEquivalent(a: IntType, b: IntType): Boolean =
-      (a, b) match {
-        case (Int1Type(sa), Int1Type(sb)) =>
-          sa == sb
-        case (IntMultiType(sa, wa, _), IntMultiType(sb, wb, _)) =>
-          sa == sb && wa == wb
-        case (BitsType(wa, _), BitsType(wb, _)) =>
-          wa == wb
-        // `u1` is equivalent to `b8`
-        case (Int1Type(false), BitsType(8, _)) => true
-        case (BitsType(8, _), Int1Type(false)) => true
-        // more generally, `u{N}` is equivalent to `b{8 * N}`
-        case (IntMultiType(false, byteWidth, _), BitsType(bitWidth, _)) =>
-          bitWidth == NUM_BITS_IN_BYTE * byteWidth.width
-        case (BitsType(bitWidth, _), IntMultiType(false, byteWidth, _)) =>
-          bitWidth == NUM_BITS_IN_BYTE * byteWidth.width
-        case _ => false
-      }
   }
   /**
     * In statically typed languages, [[CalcIntType]] is often a signed 32-bit
--- a/shared/src/main/scala/io/kaitai/struct/format/EnumSpec.scala
+++ b/shared/src/main/scala/io/kaitai/struct/format/EnumSpec.scala
@@ -4,8 +4,19 @@ import io.kaitai.struct.problems.KSYParseError
 
 import scala.collection.immutable.SortedMap
 import scala.collection.mutable
+import io.kaitai.struct.Utils
+import io.kaitai.struct.datatype.DataType.IntType
+import io.kaitai.struct.datatype.DataType
+import io.kaitai.struct.datatype.DataType.BitsType1
+import io.kaitai.struct.datatype.DataType.BitsType
+import scala.util.control.NonFatal
 
-case class EnumSpec(path: List[String], map: SortedMap[BigInt, EnumValueSpec]) extends YAMLPath {
+case class EnumSpec(
+  path: List[String],
+  doc: DocSpec,
+  intType: IntType,
+  map: SortedMap[BigInt, EnumValueSpec]
+) extends YAMLPath {
   var name = List[String]()
 
   /**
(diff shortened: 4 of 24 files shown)


generalmimon:
Let me summarize the results of this PR. The syntax changes and semantics are explained in https://github.com/kaitai-io/kaitai_struct/issues/1288, so I won't repeat them here. Instead, I will focus on the observable changes in the generated code.

### Summary

1. As per https://github.com/kaitai-io/kaitai_struct/issues/358#issuecomment-365688068, enum-level `doc` and `doc-ref` are written into the generated code as docstrings: https://github.com/kaitai-io/kaitai_struct_compiler/commit/06368c78868159856c515e414e95ca4ea380447e

   The `Docstrings` and `DocstringsDocref` tests were extended to demonstrate the use of enum-level `doc` / `doc-ref` keys: https://github.com/kaitai-io/kaitai_struct_tests/commit/4bffc29c1c6695fa076fe1a06d35e72dc9769349

2. To enable full support for 32-bit/64-bit signed/unsigned integers and thereby fix errors in the tests

   - `EnumIntRangeU` ([.ksy](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/formats/enum_int_range_u.ksy), [.kst](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/spec/ks/enum_int_range_u.kst))
   - `EnumLongRangeS` ([.ksy](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/formats/enum_long_range_s.ksy), [.kst](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/spec/ks/enum_long_range_s.kst))
   - `EnumLongRangeU` ([.ksy](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/formats/enum_long_range_u.ksy), [.kst](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/spec/ks/enum_long_range_u.kst))

   in C#, Go and Zig, enums in these languages are now declared with fixed underlying type: https://github.com/kaitai-io/kaitai_struct_compiler/commit/5f1c01fe2f00a190c2a57475260a099c20ae8a01

   - Related C# issues fixed by this commit: https://github.com/kaitai-io/kaitai_struct/issues/862, https://github.com/kaitai-io/kaitai_struct/issues/92

   - Related PR in the compiler to fix the issue that `int` is only 32 bits wide on 32-bit platforms, so the enum support is then limited to 32 bits: https://github.com/kaitai-io/kaitai_struct_compiler/pull/248

     https://github.com/kaitai-io/kaitai_struct_compiler/pull/248 is made unnecessary by this PR, because https://github.com/kaitai-io/kaitai_struct_compiler/commit/5f1c01fe2f00a190c2a57475260a099c20ae8a01 ensures that we no longer use the `int` type with platform-dependent width for enums at all. However, it shows that we should probably start running our test suite in Go on some 32-bit platform as well (not only `linux/amd64`), so that we can detect similar issues in the future.

3. To avoid undefined behavior in C++11 when casting out-of-range integers to our generated enum types (see https://github.com/kaitai-io/kaitai_struct/issues/959), the C++11 generated code now always declares enums with fixed underlying type too: https://github.com/kaitai-io/kaitai_struct_compiler/commit/8988def665ab5d152fb1475434334a3bc2a031c6

   The C++98 code still declares enums without fixed underlying type (just as before), because declaring a fixed underlying type is a feature available only since C++11, as I mentioned in https://github.com/kaitai-io/kaitai_struct/issues/959.

### Changes in test results

#### C++11 - `cpp_stl_11/clang18-linux-x86_64`

The `Enum{Int,Long}Range{S,U}` tests were already passing in C++11 previously, so there is no difference in test results during a normal test run. To verify that https://github.com/kaitai-io/kaitai_struct/issues/959 has been fixed for C++11, we need to run the test suite under [UndefinedBehaviorSanitizer](https://releases.llvm.org/18.1.6/tools/clang/docs/UndefinedBehaviorSanitizer.html) (UBSan).

<details>
<summary>Running the test suite with UBSan</summary>

This is what I used to enable UBSan in [`spec/cpp_stl_11/CMakeLists.txt`](https://github.com/kaitai-io/kaitai_struct_tests/blob/300e5135e7888b565084c875175e7582c63d69f5/spec/cpp_stl_11/CMakeLists.txt):

```diff
diff --git i/spec/cpp_stl_11/CMakeLists.txt w/spec/cpp_stl_11/CMakeLists.txt
index cc5af639..1716f9c4 100644
--- i/spec/cpp_stl_11/CMakeLists.txt
+++ w/spec/cpp_stl_11/CMakeLists.txt
@@ -9,6 +9,7 @@ find_package(Iconv REQUIRED)
 # Enforce UTF-8 source files encoding for MSVC
 add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")

+set(CMAKE_BUILD_TYPE Debug)
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
@@ -69,6 +70,9 @@ target_compile_options(ks_tests PRIVATE
        >
 )

+target_compile_options(ks_tests PUBLIC -fno-omit-frame-pointer -fno-optimize-sibling-calls -O1 -fsanitize=undefined)
+target_link_options(ks_tests PUBLIC -fno-omit-frame-pointer -fno-optimize-sibling-calls -O1 -fsanitize=undefined)
+
 add_test(ks_tests ks_tests)

 enable_testing()
```

However, it turned out that if you try to run the test suite now using our Docker image [`kaitai-cpp_stl-clang18-linux-x86_64`](https://github.com/kaitai-io/kaitai_struct_docker_images/pkgs/container/kaitai-cpp_stl-clang18-linux-x86_64/828315619) (`./docker-ci -t cpp_stl -u _11 -i clang18 -r`) built according to [`src/cpp_stl/clang18-linux-x86_64/Dockerfile`](https://github.com/kaitai-io/kaitai_struct_docker_images/blob/ebc4c5022439b3579d963b7037082f2db4e8965d/src/cpp_stl/clang18-linux-x86_64/Dockerfile), the linking phase fails:

```
[  1%] Linking CXX executable ks_tests
/usr/bin/cmake -E cmake_link_script CMakeFiles/ks_tests.dir/link.txt --verbose=1
/usr/bin/c++ -g -fno-omit-frame-pointer -fno-optimize-sibling-calls -O1 -fsanitize=undefined CMakeFiles/ks_tests.dir/runtime/cpp_stl/kaitai/kaitaistream.cpp.o (...) -o ks_tests (...)
/usr/bin/ld: cannot find /usr/lib/llvm-18/lib/clang/18/lib/linux/libclang_rt.ubsan_standalone-x86_64.a: No such file or directory
/usr/bin/ld: cannot find /usr/lib/llvm-18/lib/clang/18/lib/linux/libclang_rt.ubsan_standalone_cxx-x86_64.a: No such file or directory
c++: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [CMakeFiles/ks_tests.dir/build.make:9731: ks_tests] Error 1
```

So the missing `libclang-rt-dev` package needs to be installed:

```diff
diff --git i/src/cpp_stl/clang18-linux-x86_64/Dockerfile w/src/cpp_stl/clang18-linux-x86_64/Dockerfile
index 389b6e8..8254d4a 100644
--- i/src/cpp_stl/clang18-linux-x86_64/Dockerfile
+++ w/src/cpp_stl/clang18-linux-x86_64/Dockerfile
@@ -6,6 +6,7 @@ COPY 4img/* ./
 # Install all dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
        clang \
+       libclang-rt-dev \
        make \
        cmake \
        libboost-test-dev \
```

The resulting diff of the test binary's output before and after https://github.com/kaitai-io/kaitai_struct_compiler/commit/8988def665ab5d152fb1475434334a3bc2a031c6 follows:

```diff
diff --git 1/test_out/cpp_stl_11-clang18-old/test_run-1.stdout 2/test_out/cpp_stl_11-clang18/test_run-1.stdout
index 06f086bf..d4097c88 100644
--- 1/test_out/cpp_stl_11-clang18-old/test_run-1.stdout
+++ 2/test_out/cpp_stl_11-clang18/test_run-1.stdout
@@ -1,19 +1,3 @@
-/tests/compiled/cpp_stl_11/switch_manual_enum_invalid.h:98:43: runtime error: load of value 255, which is not a valid value for type 'code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tests/compiled/cpp_stl_11/switch_manual_enum_invalid.h:98:43
-/tests/compiled/cpp_stl_11/switch_manual_enum_invalid.h:98:43: runtime error: load of value 255, which is not a valid value for type 'code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tests/compiled/cpp_stl_11/switch_manual_enum_invalid.h:98:43
-/usr/include/boost/test/tools/old/impl.hpp:107:12: runtime error: load of value 255, which is not a valid value for type 'const switch_manual_enum_invalid_t::opcode_t::code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/include/boost/test/tools/old/impl.hpp:107:12
-/usr/include/boost/test/tools/old/impl.hpp:107:20: runtime error: load of value 255, which is not a valid value for type 'const switch_manual_enum_invalid_t::opcode_t::code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/include/boost/test/tools/old/impl.hpp:107:20
-/tests/compiled/cpp_stl_11/switch_manual_enum_invalid_else.h:121:43: runtime error: load of value 255, which is not a valid value for type 'code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tests/compiled/cpp_stl_11/switch_manual_enum_invalid_else.h:121:43
-/tests/compiled/cpp_stl_11/switch_manual_enum_invalid_else.h:121:43: runtime error: load of value 255, which is not a valid value for type 'code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tests/compiled/cpp_stl_11/switch_manual_enum_invalid_else.h:121:43
-/usr/include/boost/test/tools/old/impl.hpp:107:12: runtime error: load of value 255, which is not a valid value for type 'const switch_manual_enum_invalid_else_t::opcode_t::code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/include/boost/test/tools/old/impl.hpp:107:12
-/usr/include/boost/test/tools/old/impl.hpp:107:20: runtime error: load of value 255, which is not a valid value for type 'const switch_manual_enum_invalid_else_t::opcode_t::code_enum_t'
-SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/include/boost/test/tools/old/impl.hpp:107:20

 Test module "C++/STL test units for Kaitai Struct" has failed with:
   255 test cases out of 260 passed
```

</details>

As you can see, all the errors detected by UBSan have been fixed.

#### C# - `csharp/net10.0-linux-x86_64`

* 🟢 3 tests fixed:
  * `EnumIntRangeU`
    ```diff
    -    {"status"=>"format_build_failed"}
    +    {"status"=>"passed"}
    ```
  * `EnumLongRangeS`
    ```diff
    -    {"status"=>"format_build_failed"}
    +    {"status"=>"passed"}
    ```
  * `EnumLongRangeU`
    ```diff
    -    {"status"=>"format_build_failed"}
    +    {"status"=>"passed"}
    ```

#### Go - `go/1.26-linux-x86_64`

* ➕ 1 tests added:
  * `EnumLongRangeU`
    ```rb
    {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true}
    ```
* ➖ 1 tests removed:
  * `EnumToIInvalid`
    ```rb
    {"status"=>"passed", "elapsed"=>0.0, "is_kst"=>true}
    ```

    New compile error:

    ```
    ../../compiled/go/src/test_formats/enum_to_i_invalid.go:111:33: 32768 (untyped int constant) overflows uint8
    ```

    ```console
    $ cat -n compiled/go/src/test_formats/enum_to_i_invalid.go
      (...)
        11  type EnumToIInvalid_Animal uint8
      (...)
        21  type EnumToIInvalid struct {
        22          Pet1 EnumToIInvalid_Animal
        23          Pet2 EnumToIInvalid_Animal
      (...)
       106  func (this *EnumToIInvalid) Pet2Mod() (v int, err error) {
       107          if (this._f_pet2Mod) {
       108                  return this.pet2Mod, nil
       109          }
       110          this._f_pet2Mod = true
     * 111          this.pet2Mod = int(this.Pet2 + 32768)
       112          return this.pet2Mod, nil
       113  }
    ```

    Unfortunately, this kind of error (integer type mismatch) is very common in the currently generated Go code, and using the correct underlying type for the `animal` enum just surfaced another instance. This is a more general issue and has nothing to do specifically with enums, and therefore is unrelated to this PR.

#### Zig - `zig/0.15.2-linux-x86_64`

* 🟢 3 tests fixed:
  * `EnumIntRangeU`
    ```diff
    -    {"status"=>"format_build_failed"}
    +    {"status"=>"passed"}
    ```
  * `EnumLongRangeS`
    ```diff
    -    {"status"=>"format_build_failed"}
    +    {"status"=>"passed"}
    ```
  * `EnumLongRangeU`
    ```diff
    -    {"status"=>"format_build_failed"}
    +    {"status"=>"passed"}
    ```
* 🔴 1 tests broken:
  * `EnumToI`
    ```diff
    -    {"status"=>"passed"}
    +    {"status"=>"format_build_failed"}
    ```

    This is a similar problem as in Go:

    ```
    formats/enum_to_i.zig:67:39: error: expected type 'i32', found 'u32'
            _v = @intFromEnum(self.pet_1) + 32768;
                 ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
    formats/enum_to_i.zig:67:39: note: signed 32-bit int cannot represent all possible unsigned 32-bit values
    referenced by:
        test.EnumToI: tests/enum_to_i_test.zig:21:59
        comptime: tests.zig:64:17
    ```

    ```console
    $ cat -n compiled/zig/enum_to_i.zig
        63      pub fn pet1Mod(self: *EnumToI) !i32 {
        64          if (self._m_pet_1_mod) |_v|
        65              return _v;
        66          var _v: i32 = undefined;
     *  67          _v = @intFromEnum(self.pet_1) + 32768;
        68          self._m_pet_1_mod = _v;
        69          return _v;
        70      }
    ```

    If you look at it, the Zig compiler is absolutely right here: it makes no sense to store `u32 + <const>` in an `i32` variable. So the problem is that our compiler derived the resulting type of this value instance as `CalcIntType` (which maps to `i32`) instead of a more specific type based on the types of operands (related issues: https://github.com/kaitai-io/kaitai_struct/issues/510, https://github.com/kaitai-io/kaitai_struct/issues/617). This error is therefore also unrelated to this PR.

#### Nim - `nim/2.2.10-linux-x86_64`

Although the latest CI status of the `Enum{Int,Long}Range{S,U}` tests in Nim is the following (see https://ci.kaitai.io/ and search for tests containing "range", and here's the permalink to `ci.json`: https://github.com/kaitai-io/ci_artifacts/blob/39831e9b7b0a405d8c7f5b5ff5eee937e851f717/test_out/nim/ci.json#L265-L301):

1. `EnumIntRangeS` - **passed**
2. `EnumIntRangeU` - **passed**
3. `EnumLongRangeS` - **passed**
4. `EnumLongRangeU` - **failed**

... in reality the "**passed**" status of 2. `EnumIntRangeU` and 3. `EnumLongRangeS` is a lie - they should be actually **failed** as well. I explained in detail in the commit message of https://github.com/kaitai-io/kaitai_struct_tests/commit/1785b1ce73011518485dca05061b0963a93e6378 why they passed in Nim 2.2.10 (the latest version as of this writing) until now and how I improved the tests to prevent this from happening. The relevant excerpt is:

> ```console
> (...)
> ~/kaitai_struct_tests/spec/nim# grep 'typedef.*EnumLongRangeS_Constants' /root/.cache/nim/tenum_long_range_s_d/@m..@s..@scompiled@snim@senum_long_range_s.nim.c
> typedef NI32 tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA;
> ```
>
> `NI32` is Nim's internal name for `int32`.
>
> Consequently, the assertions in `spec/nim/tenum_long_range_s.nim` (or `spec/nim/tenum_int_range_u.nim`) don't work properly, because both sides of each comparison are truncated to signed 32-bit integers:
>
> ```console
> ~/kaitai_struct_tests/spec/nim# grep '==.*EnumLongRangeS_Constants' /root/.cache/nim/tenum_long_range_s_d/@mtenum_long_range_s.nim.c
>                 if (!!(((*r__tenum95long95range95s_u218).f1 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)(IL64(-9223372036854775807) - IL64(1)))))) goto LA4_;
>                 if (!!(((*r__tenum95long95range95s_u218).f2 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)IL64(-2147483649))))) goto LA8_;
>                 if (!!(((*r__tenum95long95range95s_u218).f3 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)(-2147483647 -1))))) goto LA12_;
>                 if (!!(((*r__tenum95long95range95s_u218).f4 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)0)))) goto LA16_;
>                 if (!!(((*r__tenum95long95range95s_u218).f5 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)2147483647)))) goto LA20_;
>                 if (!!(((*r__tenum95long95range95s_u218).f6 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)IL64(2147483648))))) goto LA24_;
>                 if (!!(((*r__tenum95long95range95s_u218).f7 == ((tyEnum_EnumLongRangeS_Constants__ZSLAsl9awA9cujexdPBWfDiA)IL64(9223372036854775807))))) goto LA28_;
> ```
>
> Of course, all of these comparisons are satisfied, so the test passes. But since the EnumLongRangeS test was specifically designed to verify that signed 64-bit integer enums work correctly, it did a poor job in this case, because it was unable to detect this incorrect Nim behavior.

While digging into this, I discovered that [enums](https://nim-lang.org/docs/manual.html#types-enumeration-types) in Nim 2.2.10 are severely broken and mostly unusable for any values beyond the range of the signed 32-bit integer type. This all stems from the fact that unlike any other statically typed language we support, you cannot reliably specify a fixed underlying type of Nim enums. The underlying type is automatically decided for you by this piece of horrible code in the Nim 2.2.10 compiler:

[`compiler/ccgtypes.nim`](https://github.com/nim-lang/Nim/blob/bfeb3146d1638b39f69007a4ae5a23e23ae4e5ef/compiler/ccgtypes.nim#L932-L942)

```nim
        if firstOrd(m.config, t) < 0:
          m.s[cfsTypes].addf("typedef NI32 $1;$n", [result])
          size = 4
        else:
          size = int(getSize(m.config, t))
          case size
          of 1: m.s[cfsTypes].addf("typedef NU8 $1;$n", [result])
          of 2: m.s[cfsTypes].addf("typedef NU16 $1;$n", [result])
          of 4: m.s[cfsTypes].addf("typedef NI32 $1;$n", [result])
          of 8: m.s[cfsTypes].addf("typedef NI64 $1;$n", [result])
          else: internalError(m.config, t.sym.info, "getTypeDescAux: enum")
```

This means that if you have any negative value in the enum, the underlying type will always be `NI32` (bye bye support for `int64`-based enums with negative values). If all enum values are non-negative and they fit into 32 bits, the underlying type will always be `NI32` (bye bye support for `uint32`-based enums). If they are non-negative and fit into 64 bits, you get `NI64`. So besides `int32`-based enums, which work fully, we have support for `int64` with non-negative values, and that's it (so of course, `uint64`-based enums don't work either).

I think that this abomination really shows just how much the required underlying type as proposed in https://github.com/kaitai-io/kaitai_struct/issues/1288 is important - otherwise, you get a very poorly designed half-baked solution, as Nim so vividly demonstrates here.

So here are the changes in test results:

* 🔴 2 tests broken:
  * `EnumIntRangeU`
    ```diff
    -    {"status"=>"passed"}
    +    {"status"=>"failed",
    +     "failure"=>
    +      {"file_name"=>nil,
    +       "line_num"=>nil,
    +       "message"=>
    +        "/tests/spec/nim/tenum_int_range_u.nim(12) tenum_int_range_u\n" +
    +        "/opt/nim/lib/std/assertions.nim(45) failedAssertImpl\n" +
    +        "/opt/nim/lib/std/assertions.nim(40) raiseAssert\n" +
    +        "/opt/nim/lib/system/fatal.nim(62) sysFatal\n" +
    +        "Error: unhandled exception: /tests/spec/nim/tenum_int_range_u.nim(12, 1) `ord(r.f2) == 4294967295&apos;i64`  [AssertionDefect]\n",
    +       "trace"=>""}}
    ```
  * `EnumLongRangeS`
    ```diff
    -    {"status"=>"passed"}
    +    {"status"=>"failed",
    +     "failure"=>
    +      {"file_name"=>nil,
    +       "line_num"=>nil,
    +       "message"=>
    +        "/tests/spec/nim/tenum_long_range_s.nim(9) tenum_long_range_s\n" +
    +        "/opt/nim/lib/std/assertions.nim(45) failedAssertImpl\n" +
    +        "/opt/nim/lib/std/assertions.nim(40) raiseAssert\n" +
    +        "/opt/nim/lib/system/fatal.nim(62) sysFatal\n" +
    +        "Error: unhandled exception: /tests/spec/nim/tenum_long_range_s.nim(9, 1) `ord(r.f1) == -9223372036854775808&apos;i64`      [AssertionDefect]\n",
    +       "trace"=>""}}
    ```

As expected, these two tests now fail, which is a good thing, because it finally reflects reality.

#### Rust - `rust/1.95-linux-x86_64`

In Rust, enums are currently always `i64`-based. Therefore, the `EnumLongRangeU` has status **format_build_failed** (see https://github.com/kaitai-io/ci_artifacts/blob/a08bf924dd1c66aa4e443ba3ba010b1fca806f10/test_out/rust/ci.json#L281), because the generated format code fails to compile. [`build-2.log:25-26`](https://github.com/kaitai-io/ci_artifacts/blob/a08bf924dd1c66aa4e443ba3ba010b1fca806f10/test_out/rust/build-2.log#L25-L26) contains the specific error messages:

```
error: literal out of range for `i64`
  --> src/formats/enum_long_range_u.rs:93:13
   |
93 |             18446744073709551615 => Ok(EnumLongRangeU_Constants::LongMax),
   |             ^^^^^^^^^^^^^^^^^^^^
   |
   = note: the literal `18446744073709551615` does not fit into the type `i64` whose range is `-9223372036854775808..=9223372036854775807`
   = help: consider using the type `u64` instead
   = note: `#[deny(overflowing_literals)]` on by default


error: literal out of range for `i64`
   --> src/formats/enum_long_range_u.rs:105:50
    |
105 |             EnumLongRangeU_Constants::LongMax => 18446744073709551615,
    |                                                  ^^^^^^^^^^^^^^^^^^^^
    |
    = note: the literal `18446744073709551615` does not fit into the type `i64` whose range is `-9223372036854775808..=9223372036854775807`
    = help: consider using the type `u64` instead
```

I tried switching to the specific underlying type for each enum just like in C#, Go and Zig (see https://github.com/kaitai-io/kaitai_struct_compiler/commit/5f1c01fe2f00a190c2a57475260a099c20ae8a01). It was easy to do, but testing revealed that this breaks 25 tests: all tests with enums except `EnumLongRangeS` and `ExprEnum`. This is because `i64` is currently hardcoded by the compiler ([`RustCompiler.scala:1205-1208`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cbd2578ff11676a28cb3e5eb16139d4c2dcbe3bb/shared/src/main/scala/io/kaitai/struct/languages/RustCompiler.scala#L1205-L1208)), so we cannot easily use various underlying types of enums without removing these hardcoded `(... as i64)` casts (which would probably require changes in other places as well). Therefore, it's best to leave Rust as it is for now.

generalmimon:
I believe this PR is finished and ready for review - I don't think there is anything left to change before merging. However, I will keep it in the "draft" state, because I don't know yet whether we'll want to merge it into `master` or somewhere else, and I don't want to be merged by mistake before that is decided. AFAIK this is the first backwards-incompatible change in the .ksy language in the history of 0.x releases, so as @GreyCat pointed out in https://github.com/kaitai-io/kaitai_struct/issues/1288#issuecomment-4278287313, we shouldn't just release it in 0.12, but we should probably continue with the 0.x-compatible releases as 1.x and this should go into 2.x.

To be honest, I've never had to work with multiple release branches. I suppose there will be something like `1.x` and `2.x` Git branches, we will primarily work on `2.x` and backport some changes to `1.x` if possible. I don't know what role (if any) the `master` branch plays in this.

Update: an issue has been created for discussing 1.x / 2.x release branches - https://github.com/kaitai-io/kaitai_struct/issues/1293

